PR #290, opened 2018-03-09
NT-622 - Click on ghosted days

eiriko2303:


eiriko2303:
@reg0 Haha yes i was suprised too! 😄 
Hm, i was thinking about that too, but it did not gives any errors when i was testing it, could fixed it as a bug later if it gives any errors 👍

PR #291, opened 2018-03-09
NT-763: fix problem with too long description in Task Manager

noro5:


src/components/favorite-toggle-list/favorite-toggle-list.component.html
```diff
@@ -5,9 +5,9 @@
   <div *ngFor="let favoriteProject of favoriteSearchProjects">
     <ion-item *ngFor="let favoriteActivity of arrayFrom(favoriteProject.searchActivities.values())" class="favorite-list-item">
       <ion-label class="list-label">
-        <h5>{{favoriteProject.customerName}}</h5>
-        <h2>{{favoriteProject.projectName}}</h2>
-        <h2>{{favoriteActivity.activityDescription}}</h2>
+        <h5 class="activity-maxlength">{{favoriteProject.customerName}}</h5>
```
reg0: I would rename the class if we start using it in a different context than the activity list

PR #294, opened 2018-03-12
NT-717 - Correct displaying of comment and comment icon

eiriko2303:


src/components/activity-list/activity-multiple-entries.component.ts
```diff
@@ -81,4 +81,14 @@ export class ActivityMultipleEntriesComponent {
         event.stopPropagation();
         this.onRemoveEntries.emit(true);
     }
+
+    checkForComment(activity: ActivityModel) {
+        let hasComment = false;
```
reg0: any reason why you didn't reuse time-entry-service.getComment() ?

anyway, please move this implementation into that service, and just call it here
also checkForComment" is not a clear name, let's call it "hasDisplayableComment"

src/components/time-entry-modal/time-entry-modal.component.ts
```diff
@@ -342,6 +343,7 @@ export class TimeEntryModal {
     }
 
     getComment(atem: ActivityTimeEntryModel): string {
+        console.log(atem.entryDescription, this.projectActivity.activityDescription);
```
reg0: please remove console.log if not necessary

eiriko2303: Sorry forgot to remove it 👍

src/components/activity-list/activity-multiple-entries.component.ts
eiriko2303: Hm, i thought it was better to just return a boolean value for it, instead of check in the html the string that is beeing returned from the getComment function.

eiriko2303: Agree, changed the name now 👍

reg0: I'd still be happy if implementation of this method was in the service :) someone might want to use it somewhere in future, and it's easier to find it there than in the component

src/components/activity-list/activity-multiple-entries.component.ts
```diff
@@ -81,4 +81,14 @@ export class ActivityMultipleEntriesComponent {
         event.stopPropagation();
         this.onRemoveEntries.emit(true);
     }
+
+    hasDisplayableComment(activity: ActivityModel): boolean {
+        let hasComment = false;
+        activity.activityTimeEntries.forEach((value, map, key) => {
+            if(value.entryDescription !== "" && value.entryDescription !== activity.activityDescription) {
+                hasComment = true;
```
noro5: You can return here directly the true value and in case there are no entries found, return false in line 92 (hasComment is not needed).

reg0: or else
```typescript
return Array.from(activity.activityTimeEntries.values()).some(value => value.entryDescription !== "" && value.entryDescription !== activity.activityDescription)
```

eiriko2303: @noro5 Agree, with that it could just have returned false 👍  but i changed it to what @reg0 suggested. That was a very handy functionality, have not seen that before. So it returns true if at least one element passes the test, that is very nice 💯

src/components/activity-list/activity-multiple-entries.component.ts
reg0: @eiriko2303 even though you changed it into a one-liner, it'd still be better to have it in the service :)

eiriko2303: Ah sorry, forgot to do that, gonna fix that 😊

PR #297, opened 2018-03-12
Nt 254 - Hours for approval by PM model

NikolaiLarsen:
So the initial implementation of approval is done, there is some missing functionality, and the code needs a bit of cleanup, but please check it out to make sure we are on the same page :)


src/services/time-entry.service.ts
```diff
@@ -105,6 +106,15 @@ export class TimeEntryService {
         });
     }
 
+    getEntriesToApprovePM(startDate: Date, endDate: Date): Observable<Array<EntryForApprovalByPM>> {
+        const utcStartDate = new Date(Date.UTC(startDate.getFullYear(), startDate.getMonth(), startDate.getDate(), 0, 0, 0));
```
reg0: this is the moment when solving "an issue with timezones" becomes a must :)

NikolaiLarsen: Any suggestions on how to solve it? :)

src/models/entryForApprovalByPM.model.ts
```diff
@@ -0,0 +1,135 @@
...
+        let arr = [];
+        for (let obj of apiArray) {
+            arr.push(EntryForApprovalByPM.mapFromApi(obj));
+        }
+        return arr;
+    }
+
+    //TO-DO: Find a more fitting place for these functions
+
+    public static getProjects(entries: Array<EntryForApprovalByPM>): Array<ProjectForApproval> {
```
reg0: these methods should all be in a service, probably a new one
static methods like that (containing business logic) are bad, because in case you'd want to write unit tests (and we should, right? :)) you're not able to mock them easily

src/models/entryForApprovalByPM.model.ts
```diff
@@ -0,0 +1,135 @@
...
+       let entriesInProjectForUser = entries.filter(entry => entry.projectNo === projectNo && entry.resourceNo === resourceNo);
+       let periods = [];
+       for (let e of entriesInProjectForUser) {
+            periods.push(new PeriodForApproval(e.periodStartDate, e.periodEndDate, e.periodDescription))
+       }
+       return EntryForApprovalByPM.getUniqueValuesFromArray(periods, 'startDate');
+    }
+
+    public static getPeriodEntriesForUserInProject(entries: Array<EntryForApprovalByPM>, projectNo: string, resourceNo: string, periodStartDate: Date): Array<EntryForApprovalByPM> {
+        let formattingService: FormattingService = new FormattingService();
```
reg0: another drawback of static methods - you can't use injects

src/models/entryForApprovalByPM.model.ts
```diff
@@ -0,0 +1,135 @@
...
+        let projectArray = [];
+        for (let entry of entries) {
+            projectArray.push(new ProjectForApproval(entry.projectName, entry.projectNo, entry.customerName, entry.customerNo));
+        }
+        return EntryForApprovalByPM.getUniqueValuesFromArray(projectArray, 'no');
+    }
+
+    public static getUsersInProject(entries: Array<EntryForApprovalByPM>, projectNo: string): Array<ApprovalUser> {
+        let arr = [];
+        for (let entry of entries) {
```
reg0: it's highly advisable to use more "fluent apis" or "stream-like apis" ;) instead of verbose for loops, which are hard to read and understand at first sight, 
you could have written:

```typescript
const UNIQUE_FILTER = (x, i, a) => {
  return x instanceof Date ? a.findIndex(x => x.getTime()) == i : a.indexOf(x) == i // this is kind of tricky implementation, but if you make sure it works with dates, and leave it like that, no one would have to modify it later, then it's OK, but the usage is super easy, see below
}


entries.
  filter(entry => entry .projectNo === projectNo).
  map(entry  => new ApprovalUser(entry.resourceName, entry.resourceNo)).
  filter(UNIQUE_FILTER)
```

src/models/entryForApprovalByPM.model.ts
```diff
@@ -0,0 +1,135 @@
...
+
+    public static mapFromApiArray(apiArray): Array<EntryForApprovalByPM> {
+        let arr = [];
+        for (let obj of apiArray) {
+            arr.push(EntryForApprovalByPM.mapFromApi(obj));
+        }
+        return arr;
+    }
+
+    //TO-DO: Find a more fitting place for these functions
```
noro5: Yes, that could go to a ApprovalService

src/actions/rootActions.ts
```diff
@@ -96,6 +96,14 @@ export function setAlertFeedbackData(alertFeedbackData: AlertFeedbackData) {
     return { type: reducerStuff.SET_ALERT_FEEDBACK_DATA, alertFeedbackData };
 }
 
+export function setEntriesForApproval(entriesForApproval: Array<EntryForApprovalByPM>) {
```
reg0: I don't see a good reason why this is in redux
it should be locally loaded in the component or page

would it be shown or editable in multiple places at the same time?

src/pages/welcome-web/welcome-web.ts
```diff
@@ -74,6 +75,9 @@ export class WelcomeWebPage {
       this.periods = periods;
       this.cdr.markForCheck();
     });
+    this.timeEntryService.getEntriesToApprovePM().subscribe(res => {
```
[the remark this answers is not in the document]
NikolaiLarsen: no, its because the project manager should get a popup similar to the one appearing when you have rejected entries, to let them know that there are hours to approve

src/models/entryForApprovalByPM.model.ts
NikolaiLarsen: I agree, i will move them once i finish the reject hours stuff

src/actions/rootActions.ts
NikolaiLarsen: it would be shown/editable in the approval page, the rejection page, the entry modal, and the notification popup

NikolaiLarsen: We also want an option to show hours that has been approved in this session in a separate page

src/pages/welcome-web/welcome-web.ts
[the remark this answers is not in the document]
reg0: ok

src/models/entryForApprovalByPM.model.ts
reg0: ah, sorry, this will be more complicated in this case, because you search for unique property, let me think...

reg0: @NikolaiLarsen I think sth like that:
```typescript
const UNIQUE_FILTER = propertyName => (x, i, a) => {
  // checks if index of first found element is equal to actual index, if there are two matching items, one of them will have different index in the array
  return x[propertyName] instanceof Date ? a.findIndex(x => x[propertyName].getTime()) == i : a.indexOf(x) == i
}


entries.
  filter(entry => entry .projectNo === projectNo).
  map(entry  => new ApprovalUser(entry.resourceName, entry.resourceNo)).
  filter(UNIQUE_FILTER('no'))
```

NikolaiLarsen: i actually tried something similar to this to begin with, but because objects are compared on the memory address and not the actual values, a.indexOf(x) == i would always return true

reg0: @NikolaiLarsen hm, weird, but I won't argue, i fount it on stackoverflow ;)

ok, anyway, I'd still be happy to replace this for loop with simpler syntax as I proposed, sth like:

```typescript
const result = entries.
  filter(entry => entry .projectNo === projectNo).
  map(entry  => new ApprovalUser(entry.resourceName, entry.resourceNo));
return getUniqueValuesFromArray(result);
```

NikolaiLarsen: I guess that thread was related to primitive values? I saw something like this on stackoverflow as well and tried it out. It works for primitive values but not objects :) Okei, i will fix the syntax to make it more readable

src/actions/rootActions.ts
reg0: @NikolaiLarsen would it be much more complicated if you passed those values with @Inputs instead of keeping it in redux?

NikolaiLarsen: @reg0 I guess it wouldn't be that much more complicated, but what exactly would be the downside of using redux for this?

src/services/time-entry.service.ts
NikolaiLarsen: done :)

src/actions/rootActions.ts
reg0: @NikolaiLarsen tough question :) I just don't feel it as a good candidate to put in redux

in general, good practices is (among others) avoiding keeping data in global scope if it's not really necessary, redux is global scope and it's good for storing information used across completely separate components or information based on user's input
otherwise, if something is relevant only for one page and it's subcomponents, and if passing those information in attributes is not overly complicated, it's usually recommended

redux state is something that has to be globally maintained in the long run, some time from now, you'll be looking at root reducer and you'll wonder why is it there, where it used, and is it safe to modify, while keeping it somewhere locally in the page, will cause that it will be really only visible in its' original context

I can't tell you I'm 100% right, maybe you're right that this data will be used on couple of different components and it will become a pain to pass it around, then it's fine, but me personally, I wouldn't put it in redux from the beginning, maybe later when we find out it would simplify anything

src/models/entryForApprovalByPM.model.ts
NikolaiLarsen: done :)

src/models/entryForApprovalByPM.model.ts
NikolaiLarsen: done :)

src/models/entryForApprovalByPM.model.ts
NikolaiLarsen: done :)

src/actions/rootActions.ts
NikolaiLarsen: Okei, thanks for the info :) i will move it out of redux

NikolaiLarsen: But how would i then get the length of the list in welcome-web.ts without an additional api call to the same endpoint? The Approval page isn't related to Welcome-web in any way...

reg0: @NikolaiLarsen idea of redux is that it should store simple information (strings, numbers, simple flat collections), so you can store precisely that information there instead of keeping whole list of objects

```hasEntriesForApproval: boolean``` - stored in redux, ```@select```'ed in welcome-web


```ts
nxRedux.dispatch(entriesForApprovalChanged(updatedList))
```
this is called everywhere where list is updated, specifically when list is loaded, or when something's added to the list (won't happen?) or removed (entry approved); you'd pass the whole list to actionCreator method, reducer will know what to do, see below

```ts
case ENTRIES_FOR_APPROVAL_CHANGED:
  return action.updatedList.length > 0;
```
new value for hasEntriesForApproval in redux

reg0: that's why I was hesitating, it's kind of an edge case :D you can do it both ways, usage will be similar, amount of code will be similar, etc. both ways involve storing something in redux, so it's hard to say which one is better or worse
to sum up, my point is that I wouldn't keep whole list of objects in redux if it's not really necessary

NikolaiLarsen: I think we could make it work without having the list in redux, but then the list would have to be fetched from the API twice. Once when welcome web is loaded to check if the user should get a notification, and once when the approval page is loaded to actually show the list. Or we could store the list in a common service for both pages, but if we would have to store it globally in a service i guess redux would be just as good/bad a solution. Any thoughts on this? :)

reg0: @NikolaiLarsen storing list in a service is even worse :) 
argh, ok, I give up :D let's put the list in redux :D

noro5: @NikolaiLarsen, @reg0: I have read your discussion and still think that it would be better to get rid of this collection from redux. To solve the issue with checking if there is anything for manager to be approved, I would create a new endpoint which would simply use the *some* method to check if there is anything to be approved, so we do not load another collection in welcome page. The collection would be retrieved when it would be really needed, which is opening of the Approval screen.

NikolaiLarsen:
@noro5 Thanks for the suggestion, I have implemented it like this now, so the PR is ready for approval again :)

NikolaiLarsen:
@eiriko2303 This also happens in dev, so I think it should be handled in a separate branch. I have made a ticket

eiriko2303:
@NikolaiLarsen Agree, then this branch can maybe be merge? 💯 👌

reg0:
@noro5 ? :)
Let's do it

PR #298, opened 2018-03-13
NT-781 - Change password bug fixes and also alertFeedback component for displaying feedback

eiriko2303:


src/pages/change-password/change-password.ts
```diff
@@ -58,25 +59,30 @@ export class ChangePasswordPage {
...
+      this.navCtrl.pop();
       setTimeout(() => {
-        this.saveAnimationService.endAnimation(true);
-        this.resetForm();
+        this.ngRedux.dispatch(setAlertFeedbackData(new AlertFeedbackData(false,"","")));
+        this.ngRedux.dispatch(setShowSpinner(false));
       }, 3000);
     }, err => {
-      this.saveAnimationService.saveErrorAnimation(err.error.error);
+      this.ngRedux.dispatch(setAlertFeedbackData(new AlertFeedbackData(true,"There are incomplete required fields. Password was not changed.", "assets/images/change_password_clock.svg")))
```
reg0: you are overriding message from API, is that on purpose? Sometimes it might be meaningful, i.e. that current password was invalid

src/pages/settings/settings.html
```diff
@@ -21,4 +21,5 @@
       </ion-item>
     </ion-list>
   </div>
+  <alert-feedback *ngIf="showAlertFeedback"></alert-feedback>
```
reg0: why is this repeated? wouldn't the one from app.html be shown?

src/pages/change-password/change-password.ts
```diff
@@ -124,8 +130,11 @@ export class ChangePasswordPage {
   }
 
   resetForm() {
-    this.isPasswordSuccesfullyChanged = false;
     this.changePasswordForm.reset();
   }
 
+  showValidationMessage() {
+    this.showValidation =  true;
```
reg0: I think it should be set back to false in resetForm()

src/pages/change-password/change-password.ts
eiriko2303: Ah, sorry, changed it now 👍

src/pages/settings/settings.html
eiriko2303: True, did not remember to delete it, thanks! 😄

src/pages/change-password/change-password.ts
eiriko2303: Forgot to delete that showValidation part, it is not used at all. Thanks :)

src/pages/change-password/change-password.ts
eiriko2303: Think we need to change those messages. It says "wrong username or password".

PR #300, opened 2018-03-14
NT-780 - Possibility to register decimals with comma on web

eiriko2303:


eiriko2303:
@noro5 It should work the same way in the modal, but because of the form inside the modal, as we discussed, the modal implementation of it could be a little bit difficult. Maybe that should be a seperate branch for now? :)

noro5:
@eiriko2303: hm... depends when this task will be implemented, because it would be good to have it unified.

eiriko2303:
@noro5 Ok, then i gonna try to do it in this branch ;)

eiriko2303:
@noro5 No it should work with comma inside the modal too :)

src/components/time-input-web/time-input-web.ts
```diff
@@ -171,4 +173,9 @@ export class TimeInputWebComponent {
     let hoursDay = pae.getSummaryOfHoursForDay(date);
     return hoursDay == 0 ? "" : hoursDay.toFixed(2);
   }
+
+  checkNumberInput(event) {
```
reg0: this function should not be duplicated, let's put it in some service or helper class

src/components/time-entry-modal/time-entry-modal.component.ts
```diff
@@ -344,4 +344,17 @@ export class TimeEntryModal {
     getComment(atem: ActivityTimeEntryModel): string {
         return atem.entryDescription === this.projectActivity.activityDescription ? '' : atem.entryDescription;
     }
+
+    checkNumberInput(event) {
+        if (event.charCode == 44 || event.charCode == 46) return true;
```
reg0: instead of all of these combinations with replacing commas, I would handle it in this event handler like this:

```ts
event.preventDefault();
if (event.charCode == 44 || event.charCode == 46) {
  event.target.value += '.';
} else if (event.charCode < 48 || event.charCode > 57) {
  return false;
} else {
  event.target.value += String.fromCharCode(event.charCode)
}
```

with this, comma will never appear anywhere and it wouldn't have to be handled elsewhere
of course that's only an idea, but it's worth checking, it would simplify those things

reg0: perhaps some extra handling would be necessary for backspace and delete keys?

src/components/time-input-web/time-input-web.ts
eiriko2303: Agree, I add it to the time-entry.service :)

src/components/time-entry-modal/time-entry-modal.component.ts
eiriko2303: Hm, interesting situation. I was trying it now, just as you added it, but that did delete the value in the input field, but gonna try to make it into one function as you suggested. 👍

eiriko2303: Ok, so i tried to add this implementation that you was talking about, but did not make it work yet, but i have moved the checkNumberInput function to the timeEntryService

eiriko2303: And i also needed to add a check if the user direct type in comma or dot if the field is empty.

reg0:
@eiriko2303 
ok, I think it's good enough :D I know I'm overreacting here :D 
two things noticed: somehow unexpected things happen when you input only a dot or comma (no digits)
when you're on web, and put it in the grid and blur out the input, nothing happens - probably as if the input was empty
on mobile, if you type only comma from the custom keyboard, and click done, also it handles it as if input was blank

I'm not sure if it's worth fixing, as this is quite stupid behavior to expect from users, and no good workaround exists (instead of replacing it with 0), but maybe you have some idea for a quick fix? otherwise I'd say it's fine

and the second thing is that in the grid, when you enter ".3" or "2," and blur out, it saves properly, but it takes quite a long time before the value in grid is refreshed and shown properly - do you know why it's so long?

eiriko2303:
@reg0 
Haha, no problem, its just good that we have this PR so that we can make the best solution ;)

For the first question: i gonna check that out and see if there are something todo there. 
Yeah, but i think that is something that should be handled in a seperate branch cause that is a different input field that the one i use.

second: hm, i think it has been like that since we implemented the page, or you mean the waiting period before the hours get the correct format: like 3 -> 3.00

eiriko2303:
@reg0 Ok, so I fixed that the field in the grid is beeing blanked out on blur if the field just includes comma or dot.

PR #303, opened 2018-03-16
NT-793 - Fixing up in marking for current day

eiriko2303:
Seems like the easy click on ghosted days PR was too good to be true. 
Nikolai found a bug that is happening when you click on the ghosted day. The bug is that if you click on a day that is not monday the sliding to that specific day is not happening. The slider slides to index 0 auto even though the index in the markDay function is the one you selected, so i tried to add a timeout for 100ms to check if that works. And it works well, and no issues when having slow internet, but i want you guys to add maybe some suggestions if there are any? 😄


src/components/day-picker/day-picker-circle.component.ts
```diff
@@ -50,7 +50,10 @@ export class DayPickerCircleComponent {
                 this.onMarkDay.emit(date);
             } else {
                 this.ngRedux.dispatch(setSelectedPeriodIndex(this.periodService.getIndexOfDate(date)));
-                this.onMarkDay.emit(date);
+                setTimeout(() => {
```
reg0: no surprise, I'm not a fan of this solution :D 
unfortunately I don't see nicer one, due to ngRedux limitations, it has to be a bit complicated

one that I can think of is
```typescript
dateToMarkWhenInSelectedPeriod = null; // class field

// markDay() function
this.ngRedux.dispatch(setSelectedPeriodIndex(this.periodService.getIndexOfDate(date)));
this.dateToMarkWhenInSelectedPeriod = date; // instead of emitting onMarkDay immediately


// redux subscription in constructor:
$selectedPeriodIndex.subscribe(selectedPeriodIndex => { 
  this.selectedPeriodIndex = selectedPeriodIndex;
  // other stuff

  if (this.dateToMarkWhenInSelectedPeriod) {
    this.onMarkDay.emit(this.dateToMarkWhenInSelectedPeriod);
    this.dateToMarkWhenInSelectedPeriod = null;
  }
});  
```

eiriko2303: @reg0 Thanks for the suggestion! I am trying to implement it now, but one question. Do you mean that all this should be in the day-picker-circle.component? Because the problem i have now is that the redux is allways a marking after what i am doing so when i click on the ghosted day it is not set to that date before i am clicking on another ghosted date.

reg0: @eiriko2303 yes, all of the pieces I wrote above should be in circle component

it should work properly, that's why I moved "onMarkDay.emit" to selectedPeriodIndex.subscribe - in this case, we can be sure that redux is already updated, so we are in proper period, and now we can emit markDay event, and sliding will happen on existing date, isn't that the case?
I can't see a reason why you'd have to click once more, doesn't markDay emit? or is it still too early?

eiriko2303: Hm, after trying some workarounds it seems like that the redux is beeing changed and that the datetomarkwheniselecteperiod value is null when redux is changed. so that the markday emit is happening to late. So the slider is setting the slide index to be the default one i think (that is 0). But I gonna try to fix it. I agree, it should work, it is a nice solution 👍

reg0:
@eiriko2303 OK, I think it's ready to test, I had to change couple of things - removed slide animiation in some cases when week is changed, added spinner here and there, but now it's OK I think

eiriko2303:
@reg0 Is Heroku updated with the final changes? Cause it seems like on the splitted week 9 when you click on sunday it does not work as expected. It jumps to monday, but on the other splitted weeks it works.

reg0:
@eiriko2303 I checked it again and you're right, there's sth wrong with week #9, I'll work on it

reg0:
@eiriko2303 hopefully now it's ok, pushed a fix

eiriko2303:
@reg0 Great work! Now it works 😄

PR #304, opened 2018-03-16
#nt 730

hasnik:
Added sorting capability to welcome-web page


src/pipes/sort-by-numeric-property.pipe.ts
```diff
@@ -0,0 +1,20 @@
+import { Pipe } from "@angular/core";
```
noro5: I think this pipe is not needed anymore and you can use the sortByProperty, which is an extended version. If so, then we should delete this pipe

noro5: I have noticed a strange behaviour when sorting byt Task:
image

src/pages/welcome-web/welcome-web.ts
```diff
@@ -522,11 +525,60 @@ export class WelcomeWebPage {
...
+    return paes.sort(function (paeA, paeB) {
+        let paeADateEntry = paeA.getProjectActivityDateEntry(propertyToSortList);
+        let paeBDateEntry = paeB.getProjectActivityDateEntry(propertyToSortList);
+
+        let paeATimeQuantity = paeADateEntry.activityTimeEntryModels.length === 0 ? 0 : paeADateEntry.activityTimeEntryModels[0].quantity;
+        let paeBTimeQuantity = paeBDateEntry.activityTimeEntryModels.length === 0 ? 0 : paeBDateEntry.activityTimeEntryModels[0].quantity;
+
+        let quantityDifference = paeBTimeQuantity - paeATimeQuantity;
+
+        return reverseSortOrder
```
reg0: somehow I fell that it'd more readable like that:

```ts
return quantityDifference * (reverseSortOrder ? -1 : 1)
```

src/pipes/sort-by-property.pipe.ts
```diff
@@ -2,19 +2,28 @@ import { Pipe } from "@angular/core";
...
 export class SortByPropertyPipe {
-    transform(array: any, arg: string): any {
-        if (array instanceof Array) {
-            array.sort((a: any, b: any) => {
-                if (a.hasOwnProperty(arg) && typeof a[arg] === "number" && b.hasOwnProperty(arg) && typeof b[arg] === "number") {
-                    return b[arg] - a[arg];
-                }
-                return 0;
-            });
+  transform(array: any, arg: string, reverseOrder: boolean): any {
```
reg0: reverseOrder could be an optional parameter

src/pipes/sort-by-property.pipe.ts
```diff
@@ -2,19 +2,28 @@ import { Pipe } from "@angular/core";
...
+    if (array instanceof Array) {
+      array.sort((a: any, b: any) => {
+        if (a.hasOwnProperty(arg) && typeof a[arg] === "number" && b.hasOwnProperty(arg) && typeof b[arg] === "number") {
+          let difference = b[arg] - a[arg];
+
+          return reverseOrder
```
reg0: same as above: ```difference * (reverseOrder ? -1 : 1)```

src/pipes/sort-by-numeric-property.pipe.ts
hasnik: Any steps to reproduce this order issue? I tried to do the same, but i'm unable to do that.

PR #305, opened 2018-03-18
Nt 77 - Reports page + extracted some common parts for settings pages styling

reg0:


reg0:
@eiriko2303 problem was in the backend, I fixed it, should I merge immediately and deploy, so that this could be re-tested?

reg0:
@eiriko2303 username was in the example excel file from Jonas
I was thinking that this column makes no sense at all, because the report is always only for one person, but I'd wait for Jonas' input about that

about chosing resources, this is in fact not implemented yet, I wasn't sure when it would be and how it will work exactly, what's the logic of finding one's subordinates, etc., so I left it like that for now, of course we can improve it later

PR #309, opened 2018-03-19
Nt 254-2 - Reject hours

NikolaiLarsen:


src/components/approval-more-popover/approval-more-popover.ts
```diff
@@ -0,0 +1,49 @@
...
+  }
+
+  ionViewWillEnter() {
+    const popoverContent = document.getElementsByClassName('popover-content')[0] as HTMLDivElement;
+    popoverContent.style.left = this.navParams.data.posX + 'px';
+    popoverContent.style.top = this.navParams.data.posY + 'px';
+  }
+
+  showRejectHours() {
+    //This have been implemented in my second PR
```
noro5: seems that this comment should be removed

src/components/approval-more-popover/approval-more-popover.ts
```diff
@@ -0,0 +1,49 @@
...
+    this.viewCtrl.dismiss();
```
noro5: why are we dismissing current page?

src/pages/reject-hours/reject-hours.ts
```diff
@@ -0,0 +1,63 @@
...
+@IonicPage()
+@Component({
+  selector: 'page-reject-hours',
+  templateUrl: 'reject-hours.html',
+})
+export class RejectHoursPage {
+  entry: EntryForApprovalByPM;
+  customMessage: string = '';
+  options = [
+    { message: 'Kan du legge inn en kommentar på denne oppgaven?' },
```
noro5: somehow that is strange that we will display norwegian comments in english version of the app (maybe only save the text in norwegian, but display the english label)?

src/components/approval-more-popover/approval-more-popover.ts
NikolaiLarsen: it is removed in PR #310 it was just to let you guys know ;)

src/pages/reject-hours/reject-hours.ts
NikolaiLarsen: Bodil wanted it to be this way as it would be a bit strange if my leader sent me a message in engllish instead of norwegian, we plan to implement one set of comments for each language, but that would be in a separate ticket

src/components/approval-more-popover/approval-more-popover.ts
noro5: Yes, the comment was very usefull, just thought that it should be deleted in this one as this PR is about rejecting hours

noro5: Yes, that comments were very useful, just thought that as this is the PR where it should be removed.

PR #310, opened 2018-03-21
Nt-254-3 - Approve all entries in period

NikolaiLarsen:


src/services/time-entry.service.ts
```diff
@@ -106,18 +106,52 @@ export class TimeEntryService {
...
             const arr = EntryForApprovalByPM.mapFromApiArray(res);
             return arr;
         });
     }
 
-    approveLinePM(entryNo: number) {
-        return this.http.post(`${environmentToUse.apiIP}:${environmentToUse.apiPort}/api/approveLinePM/${this.user.country}/${this.user.no_}/${entryNo}`, {}).map(res => {
-            this.ngRedux.dispatch(deleteEntryForApproval(entryNo));
-            return res;
+    approveLinePM(entryNo: number): Observable<TimeEntryServiceStatus> {
```
noro5: Please move this function to the ApprovalService, so we have it at one place

PR #318, opened 2018-03-26
NT-254 - bugfix after making request non-transactional

NikolaiLarsen:


src/pages/approval/approval.ts
```diff
@@ -118,6 +119,16 @@ export class ApprovalPage {
       this.handleSaveSuccess();
     }, err => {
       this.handleSaveError(err.error.error);
+      let failedEntries = err.error.entries;
+      let idsToRemove = ids.filter(id => {
+        const index = failedEntries.findIndex(failedEntry => {
```
reg0: maybe you could simplify it?

```return !failedEntries.some(failedEntry => failedEntry.id === id);```

or even better:
```ts
let failedIds = failedEntries.map(it => it.id);
let idsToRemove = ids.filter(id => failedIds.indexOf(id) < 0)
```

NikolaiLarsen: Good idea, fixed it now :)

PR #319, opened 2018-03-27
Nt 253 - Approval for line managers

NikolaiLarsen:


NikolaiLarsen:
@eiriko2303 thats strange, it worked fine locally, but i have pushed some changes that should fix the issue

src/pages/approval/approval-lm/approval-lm.ts
```diff
@@ -0,0 +1,121 @@
...
+    }
+  }
+
+  approveAllEntriesInPeriodForUser(resourceNo: string, startDate: Date) {
+    this.saveAnimationService.beginAnimation();
+    let ids = this.getPeriodEntriesForUser(resourceNo, startDate).map(entry => entry.Id);
+    this.approvalService.approveLinesPM(ids, false).subscribe(res => {
+      this.handleSaveSuccess.emit(true);
+    }, err => {
+      this.handleSaveError.emit(err.error.error);
```
reg0: you could have used formattingService.formatErrorMessage()
it was meant to get rid of those error.error.error things

src/pages/approval/approval.ts
```diff
@@ -93,55 +62,35 @@ export class ApprovalPage {
...
   }
 
-  getPeriodSumOfHoursForUserInProject(projectNo: string, resourceNo: string, startDate: Date): number {
-    const arr = this.getPeriodEntriesForUserInProject(projectNo, resourceNo, startDate);
-    return this.approvalService.getSumOfHours(arr);
-  }
-
-  setOpenPeriod(projectNo: string, resourceNo: string, startDate: Date) {
-    if (this.isPeriodOpen(projectNo, resourceNo, startDate)) {
+  setOpenPeriod(obj: any) {
```
reg0: maybe some better name here? setOpenPeriod is a bit confiusing
setOpenSection? setOpenBlock ?

src/services/approval.service.ts
```diff
@@ -103,4 +127,10 @@ export class ApprovalService {
         const body = {reason: reason};
         return this.http.post(`${environmentToUse.apiIP}:${environmentToUse.apiPort}/api/rejectEntry/${this.user.country}/${this.user.no_}/${entryNo}`, body);
     }
+
+    getWeekName(description: string) {
+        let parts = description.split('/');
+        let name = parts[0]  + (parts[1].split(' ')[1] ? '-' + parts[1].split(' ')[1] : '');
```
reg0: does it handle split weeks properly? especially 3-parts weeks?
I think some code from periodService.getWeekNumber should be reused here, or both of them could be refactored to have some speareted reusable logic

src/pages/approval/approval-lm/approval-lm.ts
NikolaiLarsen: done :)

src/pages/approval/approval.ts
NikolaiLarsen: done :)

src/services/approval.service.ts
NikolaiLarsen: That was my first thought as well when implementing this, but the problem is that the logic in periodService depends on that we also have the description of the previous period, which is not always the case for the approval page. There could for example be only one period for approval at the time. So to check if the period in question is number three in a split week i guess we would have to do a call to the backend or change the endpoint for getting entries for approval so that the week number is the right one when it is sent from the backend. Or does any other solution come to mind?

reg0: @NikolaiLarsen 
periodService.getWeekNumber() works for period indexes in redux
maybe you could match PeriodForApproval with PeriodModel-s in redux using startDate?

NikolaiLarsen: @reg0 Thats a possible solution, but can we assume that the periods for approval are allways present in the PeriodModel array in redux?

reg0: I'm not sure, can't we? ah, maybe - newly hired manager wouldn't have periods from the past assigned, right?
yup, so unfortunately it seems that getting it from the backend is the only way

NikolaiLarsen: @reg0 Okei, so what should we do then, create a new endpoint for getting the previous period description, or do the changes to the description before it is sent in the first place?

reg0: @NikolaiLarsen 
I would say "do the changes to the description before it is sent", I will have to reuse it in history page anyway, so there's no point in duplicating it again, and also, it's more optimal performance-wise, than creating separate endpoint and call it for every period separately

NikolaiLarsen: @reg0 Okei, i will do the changes in the backend, and keep this in the frontend. Thx :)

PR #322, opened 2018-03-27
NT-255 - Edit mode for Approval

eiriko2303:


eiriko2303:
@NikolaiLarsen Should be fixed now with the latest commit 😄

src/services/time-entry.service.ts
```diff
@@ -287,6 +294,10 @@ export class TimeEntryService {
                 this.user.country + '/' +
                 this.user.no_ + '/' +
                 entryNo, body).subscribe(res => {
+                    if(managerTypeOfEditer > 0 && res.entry.ExpectedApprover !== this.user.no_) {
```
noro5: In my opinion TimeEntryService should handle only update of a TimeEntry and should not care about managers entries for approval (no managerTypeOfEditer needed). I think that this should be handled in approval-more component, when the modal is dismissed. It would require to have a separate endpoint (which we might re-user on some other place), which would check if the entryForApproval is still for this manager.

src/components/approval-more-popover/approval-more-popover.ts
```diff
@@ -41,8 +66,91 @@ export class ApprovalMorePopoverComponent {
...
+    timeEntryModal.onDidDismiss(info => {
+      if (info && info.shouldReload) {
+        this.updateManagerLists(data);
+      }
+    });
+  }
+
+  private updateManagerLists(data) {
+    this.http.get(`${environmentToUse.apiIP}:${environmentToUse.apiPort}/api/timeEntryRole/${this.user.country}/${this.user.no_}/${this.entry.Id}/${this.entry.resourceNo}`).subscribe(res => {
+      if(res.Role === "None") {
```
reg0: there's way too much code duplicated here, how about this:

```ts
const dispatchDeleteFunc = this.isPM ? deleteEntryForApprovalPM : deleteEntryForApprovalLM;
if (res.Role !== "None") {
    const shouldChangeManagerType = this.isPM ? res.Role === "LM" : res.Role === "PM";
    this.updateEntryValues(data);
    if (shouldChangeManagerType) {
        this.addEntryToList(this.isPM ? ManagerType.LM : ManagerType.PM, this.entry);
    } else {
        this.updateEntryInList();
    }
}
this.ngRedux.dispatch(dispatchDeleteFunc(this.entry.Id));
this.cdr.markForCheck();
```

src/components/time-entry-modal/time-entry-modal.component.ts
```diff
@@ -48,19 +42,30 @@ export class TimeEntryModal {
     currentInputValue: number;
     isOpen: boolean;
 
+    //Approval edit part
+    managerTypeOfEditer: number;
```
reg0: why is this a number and not an enum type?

src/components/time-entry-modal/time-entry-modal.component.ts
```diff
@@ -81,6 +86,11 @@ export class TimeEntryModal {
             this.worktypes = worktypes
         });
 
+        //Approval edit part
+        if (this.managerTypeOfEditer > 0) {
```
reg0: treating it as numbers is not very readable

src/services/editModePM.service.ts
```diff
@@ -0,0 +1,45 @@
...
+    allProjects: Map<string, ProjectModel>;
+
+    constructor() {
+        this.allProjects$.subscribe(allProjects => {
+            this.allProjects = allProjects;
+        });
+    }
+
+    getProjectsForCustomer(customerName: string): Array<any> {
+        let projectsForCustomer = [];
```
reg0: please replace it with stream-like api

```ts
    this.allProjects.filter(project => project.customerName === customerName)
    .map(project => {
        return {
            projectNo: project.projectNo,
            projectName: project.projectName
        };
    });
```

src/services/editModePM.service.ts
```diff
@@ -0,0 +1,45 @@
...
+                    projectName: project.projectName
+                }
+                projectsForCustomer[i++] = projectInfo;
+            }
+        });
+        return projectsForCustomer;
+    }
+
+    getActivitiesForProject(projectNo: string): Array<string> {
+        let activitiesForProject = [];
```
reg0: same as above

src/services/time-entry.service.ts
```diff
@@ -106,27 +107,29 @@ export class TimeEntryService {
...
-                    this.ngRedux.dispatch(removeRejectedEntry(entryNo));
-                } else {
-                    activity.activityTimeEntries.forEach((value, key, map) => {
-                        value.quantity = quantity;
-                        value.entryDescription = comment;
-                        value.worktype = worktype;
-                        value.resetNotApproved();
-                        this.ngRedux.dispatch(removeRejectedEntry(key));
-                    });
+                if (managerTypeOfEditer === 0) {
```
reg0: number

src/components/time-entry-modal/time-entry-modal.component.ts
noro5: maybe simply name it: userRole

src/models/enums/managerType.enum.ts
```diff
@@ -0,0 +1,5 @@
+export enum ManagerType {
```
noro5: That enum I would name : UserRole

src/components/approval-more-popover/approval-more-popover.ts
eiriko2303: This looks much better, but the problem is that this does not update the current list if the value is just going to be updated with a new task or if the project is still on the line manager list (if it was on that list before)

src/components/time-entry-modal/time-entry-modal.component.ts
eiriko2303: My bad should be enum type 👍

src/components/time-entry-modal/time-entry-modal.component.ts
eiriko2303: Gonna change it to this: this.managerTypeOfEditer !== ManagerType.NONE

src/services/editModePM.service.ts
eiriko2303: allProjects is of type Map<string, ProjectModel> so you cant use the method filter on it. And this will just return a single project?

src/services/editModePM.service.ts
eiriko2303: Same comment as the one above?

src/services/time-entry.service.ts
eiriko2303: Gonna fix it 👍

src/services/editModePM.service.ts
reg0: then

```ts
    Array.from(this.allProjects.values())
    .filter(project => project.customerName === customerName)
    .map(project => {
        return {
            projectNo: project.projectNo,
            projectName: project.projectName
        };
    });
```

src/services/editModePM.service.ts
reg0: same reply as above :)

src/services/editModePM.service.ts
reg0: why single project? do you expect single one?

eiriko2303: I thought that this function was returning the first object where the customer is the customerName parameter. But i tried this now and it returns undefined to the modal. I change it a bit to this:
let projectsForCustomer = [];
        Array.from(this.allProjects.values())
        .filter(project => project.customerName === customerName)
        .map(project => {
            projectsForCustomer.push({
                projectNo: project.projectNo,
                projectName: project.projectName
            });
        });
        return projectsForCustomer;

And then it works, is that good enough or?

src/models/enums/managerType.enum.ts
eiriko2303: Agree gonna change it.

src/components/approval-more-popover/approval-more-popover.ts
reg0: ok, I see, so maybe like that:


```ts
const dispatchDeleteFunc = this.isPM ? deleteEntryForApprovalPM : deleteEntryForApprovalLM;
if (res.Role === "None") {
    this.ngRedux.dispatch(dispatchDeleteFunc(this.entry.Id));
    this.cdr.markForCheck();
} else {
    const shouldChangeManagerType = this.isPM ? res.Role === "LM" : res.Role === "PM";
    this.updateEntryValues(data);
    if (shouldChangeManagerType) {
        this.ngRedux.dispatch(dispatchDeleteFunc(this.entry.Id));
        this.addEntryToList(this.isPM ? ManagerType.LM : ManagerType.PM, this.entry);
    } else {
        this.updateEntryInList();
    }
}
```

still a bit complicated, but at least without duplicated code

src/services/editModePM.service.ts
eiriko2303: Fixed

src/services/editModePM.service.ts
eiriko2303: Fixed

eiriko2303:
Ok @reg0 and @noro5 the fixes should have been added know, thanks for the help 👍

PR #325, opened 2018-04-03
NT-826 - Modifications to taskmanager

eiriko2303:
…ustments from Bodil, but fine for now.


src/components/search-engine/search-activity/search-activity.component.html
```diff
@@ -1,8 +1,8 @@
 <ion-list-header class="list-header" (click)="toggleExpanded(activityContainer)">
   <h6>{{searchProject.val.customerName}}</h6>
-  <h2>{{searchProject.val.projectName}}</h2>
+  <br>
```
reg0: why ```<br>``` and not margin-top on following ```<h2>```?

reg0:
not sure what was supposed to be changed here, because there's no description in jira, but anyway, I posted a screenshot and those lines looks weird, there's some bit of white background partially below the search input, lines are not aligned and look weird

eiriko2303:
@reg0 I was working with Bodil on this task so she was commenting when i was doing those changes. But yes that searchfield line should definitely not have the no-result-background in the back of it, gonna fix that 👍

PR #326, opened 2018-04-03
NT-777 - Make sure selected period doesn't change when navigating from menu to welcome page

NikolaiLarsen:


src/pages/welcome/welcome.ts
```diff
@@ -148,7 +148,9 @@ export class WelcomePage implements OnInit {
       this.favouriteProjects = favs;
     });
     this.getDateArray(() => {
-      this.periodService.setSelectedPeriodToCurrentPeriod();
+      if (this.selectedPeriodIndex === this.currentPeriodIndex) {
```
reg0: why do we need this at all?
```periodService.setSelectedPeriodToCurrentPeriod()``` only sets ```selectedPeriodIndex``` and ```currentPeriodIndex``` to current week, but if they are already equal, then it won't change anything, will it?

Comment (NikolaiLarsen) [reply src/pages/welcome/welcome.ts]: ïts for setting the indexes initially when both are null. I really just looked at how this problem was solved on the desktop version, and copied the solution.

Comment (reg0) [reply src/pages/welcome/welcome.ts]: @NikolaiLarsen so maybe let's rewrite the condition so that it does what it's meant to do :)

Comment (NikolaiLarsen) [reply src/pages/welcome/welcome.ts]: You're right, i have fixed it now :)

File: src/pages/welcome-web/welcome-web.ts [hunk 2]
```diff
@@ -453,7 +453,7 @@ export class WelcomeWebPage {
     this.periodService.getPeriods(this.user.no_, this.user.country).subscribe(periods => {
       this.ngRedux.dispatch(setPeriods(periods));
       //Makes sure that we only load current period when the page is initialized, otherwise load selected period
-      if (this.selectedPeriodIndex === this.currentPeriodIndex) {
+      if (this.selectedPeriodIndex === 0 && this.currentPeriodIndex === null) {
         this.periodService.setSelectedPeriodToCurrentPeriod();
```
noro5: I would suggest to rename that method setSelectedPeriodToCurrentPeriod to something like: initializeCurrentAndSelectedPeriod, because based on the condition, it seems like we are setting selectedPerdioxIndex to null.

NikolaiLarsen: Ok, done :)

PR #328, opened 2018-04-03
NT-825 - Hiding previous button on start of period array and next on end of period array

eiriko2303:


src/components/week-picker/week-picker.scss
```diff
@@ -7,6 +7,11 @@ week-picker {
     .weekname {
         font-size: 12px;
     }
+
+    .hide-element {
+        z-index: -99999;
```
reg0: @eiriko2303 why not visibility: hidden?

eiriko2303: Ah, did not think of that solution, thanks! 👍  the problem was that i could not use display none because then the week-picker will be moved, but with visiblity it does the same 😄

PR #329, opened 2018-04-04
NT-802, NT-803 and NT-824 - Small design adjustments

eiriko2303:


reg0:
@eiriko2303 I approved, but should this button in change password form, be also that pale green on web version? there's different background there and it's not very visible

PR #330, opened 2018-04-04
ios dist

NikolaiLarsen:
This will currently only work in staging as a https connection is required to download the .plist file


src/pages/welcome/welcome.ts
```diff
@@ -132,6 +134,40 @@ export class WelcomePage implements OnInit {
...
+          {
+            text: 'Download',
+            handler: () => {
+              window.location.href = link;
+              let alert2 = this.alertCtrl.create({
+                title: 'The app is being downloaded',
+                message: 'Please close your browser and approve the developer account "Sopra Steria AS" under "Settings > General > Profiles and Devices"',
+                buttons: [
+                  {
+                    text: 'Okei',
```
reg0: "Okei" is not English :) should be OK or Okay :D

NikolaiLarsen: Hehe, fixed it now, thanks :)

NikolaiLarsen:
@reg0 I have fixed it now so that the popup only shows on the initial load of the welcome page

PR #331, opened 2018-04-04
NT-827 - Fixing agressive popup

eiriko2303:


src/pages/welcome-web/welcome-web.ts
```diff
@@ -129,7 +137,10 @@ export class WelcomeWebPage {
       } else if (rejectedNotFoundCallback) {
         rejectedNotFoundCallback();
       } else {
-        this.navigateToApproval(true);
+        console.log(this.showApprovalPopup);
```
NikolaiLarsen: Remove this maybe? :)

eiriko2303: My mistake, removed it now :)

src/pages/approval/approval.ts
```diff
@@ -43,6 +44,24 @@ export class ApprovalPage {
...
+    this.showApproveReminder(true);
+  }
+
+  showApproveReminder(showPopup: boolean, rejectedNotFoundCallback?: Function) {
+    this.approvalService.getEntriesToApprovePM().subscribe(approvalEntriesPM => {
+      this.approvalService.getEntriesToApproveLM().subscribe(approvalEntriesLM => {
+        if((approvalEntriesPM && approvalEntriesPM.length > 0) || (approvalEntriesLM && approvalEntriesLM.length > 0)) {
+          if (showPopup) {
+            this.popupService.openPopup('<img src="assets/images/unhappy-illustration.svg" class="approval-hours-image"/><br><span class="approval-hours-title">You still need to approve hours</span><span class="approval-hours-line"></span>', '<span class="approval-hours-text">Continue approving hours now,<br>you are behind</span>', 'OK, show me', 'I´ll do it later', () => {
+              this.navCtrl.push(ApprovalPage);
```
reg0: I don't understand why you're pushing approval page within approval page?
and I see that it's being called on ionViewWillLeave() but it looks very confusing

src/pages/welcome-web/welcome-web.ts
```diff
@@ -129,7 +137,9 @@ export class WelcomeWebPage {
       } else if (rejectedNotFoundCallback) {
         rejectedNotFoundCallback();
       } else {
-        this.navigateToApproval(true);
+        if(!this.showApprovalPopup) {
```
reg0: I also don't understand why ```navigateToApproval()``` is called when ```showApprovalPopup``` is false, assuming that this method is meant to set ```showApprovalPopup``` back to true if needed, 

looks overcomplicated

src/pages/approval/approval.ts
eiriko2303: This popup is a different one from the one when you first open worktime, that is because this one should be shown when you leave (ionviewwillleave) the approval page because we want the managers to go back to the approval page for continue working on the approving part.

src/pages/welcome-web/welcome-web.ts
eiriko2303: The whole redux stuff with the showApprovalPopup is just for making the approval popup show up one single time at opening of the time, but i gonna change the logics so that it is true before it is set to false (because then it has been displayed ones). That would be much better i think @reg0?

eiriko2303:
@noro5 Yep thats correct, thanks for notifing, Jonas wanted to change the text so :)

PR #335, opened 2018-04-05
Added autofocus on search button click. Fixed submit week button display

slomianadam:


eiriko2303:
@slomianadam Just did a quick test on web and iOS. On web it seems to work (probably on Android too), but on iOS it did not. Did a quick code review and it looks good, but it looks like you have another formating of the documents on your pc than us, so then this branch could add some merge conflicts for other branches. Currently in this branch there are a big change to most part of the welcome-web page, and i dont think that you have used that page?

reg0:
@eiriko2303 those changes in welcome-web look like just adding indentation which was missing
probably it was auto-formatting

slomianadam:
Ok, I will take a look. It is strange, because mostly, I was working on IPhone X emulator.

eiriko2303:
@reg0 aha ok

hasnik:
@eiriko2303 Actually I advised @slomianadam to use auto-formatting in welcome-web because I've seen there were some issues with formatting.

reg0:
@eiriko2303 please re test on an ios device
unfortunately, ios seems to do it on purpose, it's a known issue with programiatically setting focus

PR #338, opened 2018-04-09
NT-791 - Small changes to design on settings, change-password and device management page

eiriko2303:


src/pages/my-devices/my-devices.scss
```diff
@@ -56,6 +56,14 @@ page-my-devices {
         position: relative;
         float: right;
     }
+
+    &.worktime-standard-page {
+        .web {
+          .contents-container {
+            margin-left: 40px;
```
reg0: ```.worktime-standard-page``` was supposed to be used the other way around - you should define those styles in a common place - in app.scss, there's section for that class, and in case you need to do something "non-standard", you would override it in pages' scss files

PR #346, opened 2018-04-12
NT-833 - Total hours and billable hours for period implementation in Approval

eiriko2303:


eiriko2303:
@NikolaiLarsen Then the thing that we talked about was fixed ;) The getTotalSummaryOfHoursForPeriod is then moved to the approval service :)

src/pages/approval/approval-lm/approval-lm.ts
```diff
@@ -106,11 +110,15 @@ export class ApprovalLmComponent implements OnInit{
   }
 
   getSumOfBillableHoursInPeriodForUser(resourceNo: string, startDate: Date): number {
-    const entries = this.getPeriodEntriesForUser(resourceNo, startDate).filter(entry => entry.chargeable);
-    return this.approvalService.getSumOfHours(entries);
+    if(this.periodHoursSummary && this.periodHoursSummary.length > 0) {
+      let totalBillableHours = this.periodHoursSummary.find(summaryObject => summaryObject.resourceNo === resourceNo && summaryObject.periodStartDate.getTime() === startDate.getTime()).billableHours;
```
reg0: it can throw undefined error when .find() method finds nothing, when it tries to access .billableHours

src/pages/approval/approval.ts
```diff
@@ -39,6 +41,12 @@ export class ApprovalPage {
           }
          });
        });
+
+       this.approvalService.getPeriodHoursSummary().subscribe (result => {
+        this.periodHoursSummary = result;
+        console.log(this.periodHoursSummary);
```
reg0: please remove console log

src/services/approval.service.ts
```diff
@@ -145,4 +145,19 @@ export class ApprovalService {
...
+    getPeriodHoursSummary(): Observable<Array<PeriodHoursSummary>> {
+        return this.http.get(`${environmentToUse.apiIP}:${environmentToUse.apiPort}/api/periodHoursSummary/${this.user.country}/${this.user.no_}`).map(res => {
+            const arr = PeriodHoursSummary.mapFromApiArray(res);
+            return arr;
+        });
+    }
+
+    getTotalSummaryOfHoursForPeriod(resourceNo: string, startDate: Date, periodHoursSummary: Array<PeriodHoursSummary>): number {
+        if(periodHoursSummary && periodHoursSummary.length > 0) {
+        return periodHoursSummary.find(summaryObject => summaryObject.resourceNo === resourceNo && summaryObject.periodStartDate.getTime() === startDate.getTime()).totalHoursRegistered;
```
reg0: the same problem, possible undefined when .find() returns nothing

src/pages/approval/approval-lm/approval-lm.ts
reg0: plus, this piece of code is nearly the same as method in approval service for getting total hours

wouldn't it be better, to change that service method so that it returns single PeriodHoursSummary object and here, only access proper field out of it?

eiriko2303: Yeah thats true, changed it now! :) Added a fix for the undefined.

src/pages/approval/approval.ts
eiriko2303: Removed it now :)

src/services/approval.service.ts
eiriko2303: Added a fix for the undefined.

src/services/approval.service.ts
```diff
@@ -153,11 +153,11 @@ export class ApprovalService {
         });
     }
 
-    getTotalSummaryOfHoursForPeriod(resourceNo: string, startDate: Date, periodHoursSummary: Array<PeriodHoursSummary>): number {
+    getSummaryOfHoursForPeriodObject(resourceNo: string, startDate: Date, periodHoursSummary: Array<PeriodHoursSummary>): PeriodHoursSummary {
         if(periodHoursSummary && periodHoursSummary.length > 0) {
-        return periodHoursSummary.find(summaryObject => summaryObject.resourceNo === resourceNo && summaryObject.periodStartDate.getTime() === startDate.getTime()).totalHoursRegistered;
+            return periodHoursSummary.find(summaryObject => summaryObject.resourceNo === resourceNo && summaryObject.periodStartDate.getTime() === startDate.getTime());
```
reg0: this can still return undefined instead of null

src/pages/approval/approval-lm/approval-lm.ts
```diff
@@ -161,6 +157,7 @@ export class ApprovalLmComponent implements OnInit{
   }
 
   getTotalSummaryOfHoursForPeriod(resourceNo: string, startDate: Date): number {
-    return this.approvalService.getTotalSummaryOfHoursForPeriod(resourceNo, startDate, this.periodHoursSummary);
+    let summaryHoursPeriodObject = this.approvalService.getSummaryOfHoursForPeriodObject(resourceNo, startDate, this.periodHoursSummary);
+    return summaryHoursPeriodObject === null ? 0 : summaryHoursPeriodObject.totalHoursRegistered;
```
reg0: it'd be better to check ```!summaryHoursPeriodObject``` instead of comparing to null, it will work then for nulls, undefined and 0 will also be ok

eiriko2303: Yep i changes that so that it checks both if it is undefined and null.

src/services/approval.service.ts
eiriko2303: Yep i added a fix that checks both undefined and null.

PR #347, opened 2018-04-12
NT-777 - fixed date switching on navigation

NikolaiLarsen:


src/components/top-menu-mobile/top-menu-mobile.ts
```diff
@@ -226,13 +245,15 @@ export class TopMenuMobileComponent implements OnInit {
       if (!dateEl) {
         this.currentDate.setHours(this.currentDate.getHours() - 1);
```
noro5: Why we have this subtraction here?

PR #349, opened 2018-04-12
NT-867 - fixed number of decimals in approval summaries

NikolaiLarsen:


eiriko2303:
@reg0 Aha yes Kamil, thats correct, I forgot that 👍 The method is called formatToTwoDecimals. Please change it to this.

NikolaiLarsen:
@reg0 @eiriko2303 Fixed it now :)

reg0:
@NikolaiLarsen does it make sense like that? take number, format it and parseFloat() it back? it will practically remove formatting
either these functions should return strings and then parseFloat() is not needed
or this formatting should be applied on rendering and not where the values are returned

NikolaiLarsen:
@reg0 you are right, i changed it to string :)

PR #350, opened 2018-04-13
Changed headers web design to be smaller

hasnik:


hasnik:
@eiriko2303 I believe it's much better now

PR #356, opened 2018-04-16
NT-851 - Clearing of selected and current period index at logout + small bug at login

eiriko2303:


src/pages/welcome-web/welcome-web.ts
```diff
@@ -364,9 +364,11 @@ export class WelcomeWebPage {
   private getDatesToPeriodsMapping(periods: Array<PeriodModel>): Map<Date, PeriodModel> {
     const result = new Map();
     periods.forEach(period => {
-      period.dates.forEach((dateModel, dateKey, dates) => {
-        result.set(dateModel.date, period);
-      });
+      if(period) {
```
reg0: I can't really imagine when it could be false?

eiriko2303: I did add that check because i got an error there at login when i switched from one user to another. I can remove it, but it was an error messege. It could be something with not clearing all content at logout maybe? :)

reg0: this ```periods``` arrays comes from ```this.getDatesForPeriods([this.getSelectedPeriod()]);``` so the array looks like it's always single element, and ```getSelectedPeriod()``` would have to return nothing, 
maybe you're right, there're some specific conditions in between logout and log in when this could happen

src/services/auth.service.ts
```diff
@@ -109,7 +109,9 @@ export class AuthService {
         for (let nav of navs) {
             nav.setRoot(HomePage);
             nav.popToRoot(); 
-        } 
+        }
+        this.ngRedux.dispatch(setSelectedPeriodIndex(0));
```
reg0: probably this would be an overkill, but maybe these 2 values could be defined as constants in rootReducer.ts since they have to match default state of redux for that objects
up to you :)

src/pages/welcome-web/welcome-web.ts
eiriko2303: Its fun that we fix this small issues when probably no users would switch between users 😄

src/services/auth.service.ts
eiriko2303: Hm yeah that could be nice to add, but how would this constants be, like this? 
this.ngRedux.dispatch(setSelectedPeriodIndex(DEFAULT_VALUE_FOR_SELECTED_PERIOD));

reg0: UNINITIALIZED_SELECTED_PERIOD_INDEX and UNINITIALIZED_CURRENT_PERIOD_INDEX maybe?

but if you decide to do it, you'd have to review other parts of the code, because we already have some checks like ```if (selectedPeriodIndex == 0)``` and they should also use that constants

one more thing, @noro5 just suggested - maybe it'd be better to do that in reducer? example:
```ts

const selectedPeriodIndex = (state: number = 0, action) => {
    switch (action.type) {
        case SET_SELECTED_PERIOD_INDEX:
            state = action.selectedPeriodIndex;
            return state;
        case LOGOUT_USER:
            return 0;
        default:
            return state;
    }
}
```
that's how really reducers should be built, they have to react on actions, and it doesn't mean that one action can fire only one result

eiriko2303: Yeah that would be much better i think, then we dont need to change every place where selectedPeriodIndex == 0 and the same for the currentPeriodIndex :)

eiriko2303: Oh sorry, you mean that the change "selectedPeriodIndex == 0" would be done and also that the reducer method should be change?

reg0: ideally both
I would start with changing reducer, 
if it works fine -> introduce constants -> replace hardcoded 0's elsewhere in that context

eiriko2303: I was trying to changing the reducer now, but that introduce problems. I get no_ of null so it seems like the user object is not available or something. But maybe this changes to the reducer and the introducing of this constant should be added in a seperate branch that I could make?

eiriko2303:
@NikolaiLarsen @reg0 @noro5  I think the problems is fixed now, but it was not easy to test that the current date was selected at login as Nikolai mentioned so feedbacks could be nice :)

src/components/top-menu-mobile/top-menu-mobile.ts
```diff
@@ -275,7 +275,7 @@ export class TopMenuMobileComponent implements OnInit {
 
   //Method for marking first day in the watched period
   setFirstDate() {
-    let date = new Date(this.periods[this.selectedPeriodIndex].startDate);
+    let date = (this.isDateInPeriod(new Date(), this.periods[this.selectedPeriodIndex]) ? new Date() : new Date(this.periods[this.selectedPeriodIndex].startDate));
```
reg0: I'm pretty sure that this logic of setting today as selected date for current period was already defined somewhere, wasn't it?

eiriko2303: Yes it is i think, but the problem was that when i logged out from one user then in again with another one i was sent to monday in the current period, so not the todays date as it should do. The first time login is sending the user to the correct date on mobile, so it was a strange bug

PR #358, opened 2018-04-16
NT-873 - Added comment and hours to modal on approval part

eiriko2303:


src/components/time-entry-modal/time-entry-modal.component.scss
```diff
@@ -63,6 +63,17 @@ time-entry-modal {
             background-color: #adadad;
             border: solid 0.1rem #adadad;
         }
+
+        &-current {
```
reg0: creating such classes as "modal-current-quantity" is not always good
in this case, we have a specific context (```time-entry-modal```) and ```current-comment``` would be unique enough to make it a separate class, ```modal-``` prefix is not needed here
with what you did, it's hard to search through the code if you'd want find this class, because it's not defined as a whole in scss, so you won't find it
on the other hand, defining it like that, doesn't bring any value, except of not duplicated "current" word

imho, it'd be better if classes would be ```current-quantity``` and ```current-comment```, and css like that:
```css
time-entry-modal {
  ...
  & .current-quantity { ... }
  & .current-comment { ... }
}
```

eiriko2303: Agree, gonna change it 👍

eiriko2303:
@reg0 Changed it :)

src/components/time-entry-modal/time-entry-modal.component.scss
```diff
@@ -161,6 +150,17 @@ time-entry-modal {
         border-bottom: 1px solid #41738c;
         margin-left: 1em;
     }
+
+    .current {
```
reg0: you missed the point :) I wanted these classes to be kept in full in css, so that they are easy to find/replace if needed, see example I provided
you can leverage parent-child relation in selectors and keep full class names at the same time

at the same time, you could do similar thing with other classes that were here before on fields above, ```modal-customer-name-approval``` and others

eiriko2303: Ok so you want to remove those scss values to full css classes?

reg0: yes, in this case
```.current``` means nothing here, and does not exist as a separate class, so splitting it into parts here doesn't make sense

PR #359, opened 2018-04-16
NT-869 - Cross to remove searchfield value on taskmanager and welcome pages

eiriko2303:


src/index.html
```diff
@@ -6,7 +6,7 @@
   <meta name="viewport" content="viewport-fit=cover, width=device-width, initial-scale=1.0, minimum-scale=1.0, maximum-scale=1.0, user-scalable=no">
   <meta name="format-detection" content="telephone=no">
   <meta name="msapplication-tap-highlight" content="no">
-  <script>var env = 'dev';</script>
+  <script>var env = 'staging';</script>
```
reg0: is that change needed?

eiriko2303: Ah sorry, was testing the changes from home yesterday, nice that you did see that.

eiriko2303:
Aha, no it was not the intention with that line. It was to just hide it when the user has close the search field. So it is just visible when the searchfield is open, but yep i agree that it should be just visible when the user type something :) going to change that, thanks for the input 👍

eiriko2303:
@noro5 Now the fix for hiding the cross until user types in the field is added.

src/components/filter-search/filter-search.component.html
```diff
@@ -1,7 +1,10 @@
...
+  <img src="assets/images/search.png" class="search-icon"/>
+</button>
 <div id="filter-search" [ngClass]="{'hide': hide, 'expand': showInput}">
-  <button ion-button class="searchButton" (click)="showInputField($event)">
-    <img src="assets/images/search.png" class="search-icon"/>
-  </button>
   <input id="searchField" (click)="searchFieldClick($event)" type="text" class="searchInput" [(ngModel)]="filterTerm" [formControl]="filterControl" (blur)="hideInputField()"
     placeholder="Search in your favorites" (keypressed)="doneWriting($event)" />
+  <ng-container *ngIf="showClearInputButton()">
+    <ion-icon name="close" class="remove-searchtext-button" [ngClass]="{'hidden-clear-button': !showInput}" (click)="removeSearchTerm($event)"></ion-icon>
```
noro5: I think that you do not need anymore the !showInput condition to hide the button, as you have this ngIf on a container.

NikolaiLarsen: @noro5  Yes, its still needed. On mobile it makes sure that the x is not shown if you type something in the searchField and then close it without clearing it :)

PR #364, opened 2018-04-17
Store activity list sorting info in app state

hasnik:


src/pages/welcome-web/welcome-web.ts
```diff
@@ -608,9 +614,8 @@ export class WelcomeWebPage {
   }
 
   sortByDate(date: Date) {
-    if (this.propertyToSortList !== date) {
-      this.propertyToSortList = date;
-      this.reverseOrderToSortList = false;
+    if (this.activityListSorting.propertyToSortList !== date) {
```
reg0: I would add .getTime() here, I'm not sure if comparing Date objects directly is 100% accurate and cross-browser

src/pages/welcome-web/welcome-web.ts
```diff
@@ -633,8 +636,7 @@ export class WelcomeWebPage {
   }
 
   private resetSorting() {
-    this.propertyToSortList = 'projectActivityUsagesInLastMonth';
-    this.reverseOrderToSortList = false;
+    this.ngRedux.dispatch(setActivityListSorting('projectActivityUsagesInLastMonth', false));
```
reg0: you could also add and action like resetSorting() so that reducer sets default values, instead of duplicating them here

src/reducers/rootReducer.ts
```diff
@@ -369,7 +380,8 @@ const appStore = combineReducers<IAppState>({
     entriesForApprovalLM,
     haveShownDownloadAlert,
     approvedEntries,
-    showApprovalPopup
+    showApprovalPopup,
+    activityListSorting
```
reg0: I think you could also add this to IAppState interface, although I'm not sure what it does exactly :D

hasnik:
@reg0 Please re-review ;)

src/reducers/rootReducer.ts
reg0: sorry, missed that

reg0:
@hasnik now it's perfect :)

PR #366, opened 2018-04-18
NT-875 - Worktype visible in the edit modal on approval part

eiriko2303:
- Added worktype visible to the modal if user have it on the entry, also added some text to the current values to better showing what is quantity, comment and worktype of the text.


src/components/time-entry-modal/time-entry-modal.component.ts
```diff
@@ -93,6 +94,8 @@ export class TimeEntryModal {
             this.activitiesForProject = this.editService.getActivitiesForProject(this.projectActivity.projectNo);
             this.currentComment = this.timeEntries[0].comment;
             this.currentQuantity = this.timeEntries[0].quantity;
+            this.currentWorktype = this.timeEntries[0].worktype;
+            this.timeEntries[0].worktype !== "" ? this.showWorktype = true : null;
```
reg0: please use ternary operator properly:
```this.showWorktype = (this.timeEntries[0].worktype !== "" ? true : false);```
or replace it with normal if clause
or even better
```this.showWorktype = this.timeEntries[0].worktype !== "" ```

reg0: point is, not to use this syntax as a one-line replacement for if, it's not the same and it's less readable

eiriko2303: Aha ok, understand, added the more cleaner look now 👍

PR #367, opened 2018-04-18
NT-854 - Added scaling to report, rejected-hours and approval page

eiriko2303:


NikolaiLarsen:
@reg0 I fixed it now so that it uses common css, please re-review :)

PR #369, opened 2018-04-18
NT-832 - Showing PBR for line managers

eiriko2303:


src/pages/approval/approval-lm/approval-lm.ts
```diff
@@ -155,9 +155,36 @@ export class ApprovalLmComponent implements OnInit{
...
-    return this.formattingService.formatToTwoDecimals(summaryHoursPeriodObject ? summaryHoursPeriodObject.totalHoursRegisteredPeriod : 0);
+    return this.formattingService.formatToTwoDecimals(this.getSummaryHoursPeriodObject(resourceNo, startDate) ? this.getSummaryHoursPeriodObject(resourceNo, startDate).totalHoursRegisteredPeriod : 0);
+  }
+
+  getNormTimeInPeriodForUser(resourceNo: string, startDate: Date) {
+    return this.formattingService.formatToTwoDecimals(this.getSummaryHoursPeriodObject(resourceNo, startDate) ? this.getSummaryHoursPeriodObject(resourceNo, startDate).normTimePeriod : 0);
+  }
+
+  getSumOfVacationHoursInPeriodForUser(resourceNo: string, startDate: Date) {
+    const entries = this.approvalService.getPeriodEntriesForUserInProject(this.entriesForApprovalLM, "01", resourceNo, startDate);
```
noro5: why the project is hardcoded here? 😀

eiriko2303: Haha it is because the vacation hours should be specific added to the calculation of the PBR and the vacation projectnumber is 01 in all countries 😁

eiriko2303: Sorry i have made some changes now locally that i am going to push. I was doing it wrong. I need the totalhours registered on vacation on the period from the server not the frontend. Because this number in frontend would change if i approve some hour, but with the changes i made now it would get the total hours registered on vacation for one period from the server.

noro5: OK, for me better to not have any hard coded value :)

reg0: @eiriko2303 
"01" is correct now, but will it be always? :) or can we be sure that it won't be more different projects codes assigned to vacation?
it's good to move it to backend, because then we don't have to change anything here, but it's still hardcoded in our api

src/pages/approval/approval-lm/approval-lm.ts
```diff
@@ -155,9 +155,28 @@ export class ApprovalLmComponent implements OnInit{
...
+
+  getNormTimeInPeriodForUser(resourceNo: string, startDate: Date) {
+    return this.formattingService.formatToTwoDecimals(this.getSummaryHoursPeriodObject(resourceNo, startDate) ? this.getSummaryHoursPeriodObject(resourceNo, startDate).normTimePeriod : 0);
+  }
+
+  getSumOfVacationHoursInPeriodForUser(resourceNo: string, startDate: Date) {
+    return this.formattingService.formatToTwoDecimals(this.getSummaryHoursPeriodObject(resourceNo, startDate) ? this.getSummaryHoursPeriodObject(resourceNo, startDate).totalHoursRegisteredVacation : 0);
+  }
+
+  getPBRForPerios(resourceNo: string, startDate: Date) {
```
reg0: typo - perioDs

PR #370, opened 2018-04-19
NT-856 added more filters in reports page and logic for getting subordinates

reg0:


src/pages/welcome-web/welcome-web.ts
```diff
@@ -156,7 +156,7 @@ export class WelcomeWebPage {
   }
 
   navigateToApproval(showPopup: boolean, rejectedNotFoundCallback?: Function) {
-    if(this.user && (this.user.isProjectSupervisor || this.user.projectAdmininistrator || this.user.projectSupervisor)) {
```
NikolaiLarsen: We have found that these flags aren't always set in the right way by finance, but we have an endpoint that checks if the resource is a projectManager on any projects in db, or registered as supervisor for any resource in db. You could just store the value returned from that endpoint in a isManager field on the user object. This endpoint is called from page-list component, through approval-service

NikolaiLarsen: this new endpoint is much more reliable than the flags in the db

reg0: this place is something not related with reports, I only added method call instead of directly checking flags - I didn't changed behavior here

I fixed it on reports page on the other hand, using api endpoint you suggested, but still it needs further changes later on, but let's not do it within this branch

NikolaiLarsen: Yep, i realized that you just added a method for checking flags, just thought i should inform you about these findings ;)

src/pages/report/report.html
```diff
@@ -26,17 +16,38 @@
                 <my-date-picker [(ngModel)]="dateTo" [options]="dateToOptions" (dateChanged)="handleDateToChanged($event)"></my-date-picker>
             </div>
         </div>
+        <div class="form-item customer">
+            <ion-label>Customer</ion-label>
+            <div class="form-control">
+                <select [ngModel]="customer" (ngModelChange)="handleCustomerChanged($event)">
```
eiriko2303: Could maybe be nice to add the specific styling for this dropdowns in this branch? I have made the styling for the select dropdowns in the time-entry-modal on approval page. The class is: "select-option-wrapper" inside app.scss :)

src/pages/report/report.ts
```diff
@@ -117,28 +136,81 @@ export class ReportPage {
...
+            return result;
+        }, []).sort(compareCustomerNamesCaseInsensitive);
+    }
+
+    availableProjectsFiltered() {
+        return this.customer ? this.availableProjects.filter(it => it.customerNo === this.customer) : this.availableProjects;
+    }
+
+    reloadAvailableResources(): Promise<any> {
+        return new Promise((resolve, reject) => {
```
noro5: Maybe observable would be nicer, just in case to stay consistent to what we use.

src/pages/report/report.ts
```diff
@@ -33,6 +35,7 @@ export class ReportPage {
     dateToString: string;
     availableProjects: Array<SimpleProjectModel> = []
     project: string = "";
+    customer: string = "";
```
noro5: those names are misleading - if it is a projectNo, then I would recommend to rename it so. While CR I thought that I have here some object. Same for customer.

reg0: done

src/pages/report/report.ts
reg0: not a fan of Observables, but OK :)

PR #371, opened 2018-04-20
NT-868: refactoring AL: new model for displaying data

noro5:


src/components/activity-list/activity-list.component.ts
```diff
@@ -516,69 +487,52 @@ export class ActivityListComponent {
...
   }
 
   private sortFilteredProjectsAndActivitiesOnDate() {
-    this.filteredProjectsOnDate = this.sortByPropertyPipe.transform(this.filteredProjectsOnDate, 'numberOfUsagesInLastMonth');
-    this.filteredActivitiesOnDate.forEach((activityModels, key, map) => {
-      let sortedActivites = this.sortActivityModels(activityModels);
-      this.filteredActivitiesOnDate.get(key).splice(0, sortedActivites.length, ...sortedActivites);
-    });
+    this.filteredActivityListItems.sort(this.sortActivityListItemsByUsage);
+    this.filteredActivityListItems.sort((a, b) => this.sortRejectedEntriesPipe.sortRejectedEntriesFirst(a.activity, b.activity));
```
reg0: I'm wondering if calling .sort() twice makes sense? does second call keep order calculated in first one? I don't think that's guaranteed
probably it'd be better if you could combine it in one .sort() and put more complex logic in sorting closure

src/components/activity-list/activity-list.component.ts
```diff
@@ -516,69 +487,52 @@ export class ActivityListComponent {
...
   }
 
-  private sortActivityModels(collection: any[]): any[] {
-    let sortedByUsage = this.sortByPropertyPipe.transform(collection, 'numberOfUsagesInLastMonth');
-    let sortedByRejected = this.sortRejectedEntriesPipe.transform(sortedByUsage);
-    return Array.from(sortedByRejected);
+  private sortActivityListItemsByUsage(a: { project: ProjectModel, activity: ActivityModel }, b: { project: ProjectModel, activity: ActivityModel }) {
+    const firstItemUsage = a.project.numberOfUsagesInLastMonth + a.activity.numberOfUsagesInLastMonth;
```
reg0: why do we add it? doesn't activity's usage include project's usage?

src/components/activity-list/activity-list.component.ts
```diff
@@ -588,7 +542,7 @@ export class ActivityListComponent {
             return this.isUserScrollingDown(positions) && this.isScrollExpectedPercent(positions[1], 70);
           });
 
-        requestOnScroll$.subscribe(value => {
+        this.requestOnScrollSubscription = requestOnScroll$.subscribe(value => {
```
reg0: shouldn't you unsubscribe old one if existed?

src/pages/welcome/welcome.ts
```diff
@@ -426,16 +438,22 @@ export class WelcomePage implements OnInit {
...
   getNormTimesForPeriod(period: PeriodModel) {
-    if (period && !period.hasNormTimes) {
-      if (this.normSubscription) this.normSubscription.unsubscribe();
-      if (this.user) {
-        this.normSubscription = this.periodService.getPeriodNormTimes(this.user.no_, this.user.country, period.startDate).subscribe(periodNorms => {
-          this.ngRedux.dispatch(setNormTimes(periodNorms, this.selectedPeriodIndex));
-          //period.setNormTimes(periodNorms);
-          this.reloadList();
-        });
+    return new Observable(observer => {
```
reg0: please add method's return type

src/components/activity-list/activity-list.component.ts
noro5: Yes, Project usage includes all usages from Activities, but I want it to be sorted first by Project and then Activities in that Projects as well.

src/components/activity-list/activity-list.component.ts
noro5: If that subscription exists, I do not make new one, but if we change week and the AL is created again, I do unsubscribe in the ngOnDestroy, to prevent memory leaks.

src/pages/welcome/welcome.ts
noro5: OK.

src/components/activity-list/activity-list.component.ts
noro5: I think you are right - it is not guaranteed. Changed that.

PR #373, opened 2018-04-20
NT-849 Added Expand/Collapse side menu

slomianadam:


src/components/side-menu-web/side-menu-web.scss
```diff
@@ -36,6 +37,26 @@ side-menu-web {
...
+        .side-menu-web{
+            * {
+                display: none;
+            }
+            .menu-logo-wrapper, .menu-logo-wrapper * {
+                margin-left: -0.75rem;
+                padding: 0px;
+                height: 45px;
+                width: 18rem;
+                display: block;
```
reg0: sorry, my mistake, this is wrong here:
only ```display: block;``` should be applied to ```.menu-logo-wrapper * ```
rest should be only for ```.menu-logo-wrapper```

src/components/side-menu-web/side-menu-web.html
```diff
@@ -1,9 +1,11 @@
 
-    <div class="side-menu-web">
-      <img class="menu-logo" src="/assets/images/Worktime_logo_black.svg"/>
-      <page-list [forWeb]="true"></page-list>
-      <user-profile [isOnLightBackground]="true"></user-profile>
-      <logout-button [isOnLightBackground]="true"></logout-button>
+    <div  class="side-menu-web">
+      <div class="menu-logo-wrapper">
```
reg0: please fix indentation in following 3 lines

src/components/side-menu-web/side-menu-web.ts
```diff
@@ -22,4 +28,13 @@ export class SideMenuWebComponent {
         this.menuCtrl.close();
     }};
 
+    hideSideMenu(){
```
reg0: this method could be named ```toggleSideMenu```

src/components/side-menu-web/side-menu-web.ts
```diff
@@ -22,4 +28,13 @@ export class SideMenuWebComponent {
...
+      this.sideMenuHidden = !this.sideMenuHidden;
+      let element = document.getElementsByClassName("worktime-menu")[0];
+      if (this.sideMenuHidden) {
```
reg0: maybe this field is not needed anymore?
it'd be enough to check if that class is already there

src/components/side-menu-web/side-menu-web.ts
```diff
@@ -22,4 +28,13 @@ export class SideMenuWebComponent {
...
+        element.className += " collapsed";
```
NikolaiLarsen: you could maybe use ngClass for this instead :)

reg0: it's not that easy, because it has to be set on an element outside of this component

NikolaiLarsen: oh right, i didnt notice that

NikolaiLarsen: Then we could maybe have a sideMenuOpen field in redux and use that for ngClass?

reg0: @azslomian maybe it'd be cleaner to use Renderer
https://stackoverflow.com/questions/42809368/angular-2-adding-a-class-to-a-dom-element-when-clicked
or even https://stackoverflow.com/a/47852581/1047852

PR #375, opened 2018-04-23
NT-888 - fixed menu, filtering out my tasks for external users

NikolaiLarsen:


src/services/auth.service.ts
```diff
@@ -125,6 +127,7 @@ export class AuthService {
             const country = this.storage.getItem('country');
             const projectAdmin = this.storage.getItem('projectAdmin') == 'true';
             const projectSupervisor = this.storage.getItem('projectSupervisor') == 'true';
+            const isExternal = this.storage.getItem('isExternal') == 'true';
             const isSupervisor = this.storage.getItem('isProjectSupervisor') == 'true';
             const user = new UserModel(true, true, false, "", "", isSupervisor, false, user_name, 0, user_no, new Date(), projectSupervisor, projectAdmin, false, true, true, country);
```
noro5: Shouldn't you use the new isExternal variable when creating UserModel here?

PR #378, opened 2018-04-23
Rounding of normtime period to two decimals

hasnik:


src/components/top-menu-mobile/top-menu-mobile.html
```diff
@@ -14,7 +14,7 @@
       <div class="summary-hours-period">
         This week:
         <div class="totalhours-current-period">{{totalhoursperiod}} </div>
-        <div class="normtime-text">/ {{normtimeperiod}} h</div>
+        <div class="normtime-text">/ {{formattingService.formatToTwoDecimals(normtimeperiod)}} h</div>
```
eiriko2303: Great work! 😄 Just a suggestion, maybe it would be better to call this formattingService method inside a function and then call that function from the HTML. Like getFormattedNormTime() or something?

reg0: I agree, or else you could even create a Pipe for that and try to locate other places where it may be applied

src/components/top-menu-mobile/top-menu-mobile.ts
```diff
@@ -207,12 +207,18 @@ export class TopMenuMobileComponent implements OnInit {
...
+      this.getSelectedPeriod().getPeriodNormTime() ? this.normtimeperiod = this.calculateNormTimePeriod() : 0;
       this.selectedDateModel ? this.normtimetoday = this.selectedDateModel.normTime : 0;
       this.updateDayOrTodaytext();
     }
   }
 
+  calculateNormTimePeriod(): number {
+    let calculatedNormTimePeriod = this.getSelectedPeriod().getPeriodNormTime();
+    let formattedNormTimePeriod = this.formattingService.formatToTwoDecimals(calculatedNormTimePeriod);
+    return Number(formattedNormTimePeriod);
```
reg0: are you sure that converting it back to number type will not make it lose decimal digits information?

src/components/top-menu-mobile/top-menu-mobile.ts
```diff
@@ -207,12 +207,18 @@ export class TopMenuMobileComponent implements OnInit {
       } else {
         this.totalhourstoday = parseFloat(this.selectedDateModel.sumAllQuantities().toFixed(2));
       }
-      this.getSelectedPeriod().getPeriodNormTime() ? this.normtimeperiod = this.getSelectedPeriod().getPeriodNormTime() : 0;
+      this.getSelectedPeriod().getPeriodNormTime() ? this.normtimeperiod = this.calculateNormTimePeriod() : 0;
```
reg0: please change it to proper ternary expression:

```this.normtimeperiod = this.getSelectedPeriod().getPeriodNormTime() ? this.calculateNormTimePeriod() : 0;```

File: src/components/top-menu-mobile/top-menu-mobile.html [hunk 4]
```diff
@@ -14,7 +14,7 @@
       <div class="summary-hours-period">
         This week:
         <div class="totalhours-current-period">{{totalhoursperiod}} </div>
-        <div class="normtime-text">/ {{formattingService.formatToTwoDecimals(normtimeperiod)}} h</div>
+        <div class="normtime-text">/ {{normtimeperiod}} h</div>
```
reg0: you could have just call a function here, instead storing formatted value inside controller

```{{formatDecimal(normtimeperiod)}}```

and 
```ts
formatDecimal(n: number): string { 
    return this.formattingService.formatToTwoDecimals(n)
}
```

PR #380, opened 2018-04-23
#NT-906: Make customer, project and task selectable in Register Hours view

hasnik:


src/pages/welcome-web/welcome-web.scss
```diff
@@ -297,6 +297,13 @@ page-welcome-web {
         font-size: 14px;
         font-weight: normal;
     }
+
+    .user-selectable-text {
```
reg0: this class and styles were already defined in favorite-toggle-list.component.scss
can you move it to app.scss? because it seems that it is global and we can reuse it in other places

PR #381, opened 2018-04-23
NT-896 - make sure popup is only showed once

NikolaiLarsen:


src/components/top-menu-mobile/top-menu-mobile.ts
```diff
@@ -130,15 +130,18 @@ export class TopMenuMobileComponent implements OnInit {
...
     this.timeEntryService.getRejectedEntryDates().subscribe(rejectedDates => {
       this.datesRejected = rejectedDates;
-      if (rejectedDates !== null) {
-        if (showPopup) this.popupService.openPopup('<img class="rejected-message-image" src="assets/images/rejected-hours-message.svg"/><b><p class="rejected-message-title">Attention!</p></b>', '<span class="reject-message-hline"></span><b><p class="rejected-message-text">' + rejectedDates.length + ' entr' + (rejectedDates.length > 1 ? 'ies' : 'y') + ' require changes.</p></b>', 'OK, show me', 'I´ll do it later', () => {
-          this.goToRejected(rejectedDates);
-        });
-        else {
-          this.goToRejected(rejectedDates);
+      if (!this.popupService.haveTriggeredPopups) {
+        this.popupService.haveTriggeredPopups = true;
```
reg0: I'm not sure if this is a good solution, did you try to find out WHY they were shown twice?
soon, we might want to reuse this pop-ups for showing push notifications, and this might block them

noro5: For me it looks like that we are calling navigateToRejected twice, so I would look for root of that. There are similar problems in Activity List (called after we load: Period, Time Entries, Norm Times), so it was loaded three times, when this component was created.

NikolaiLarsen: its actually ngOnInit that is beeing called twice because of navCtrl.setRoot, and ngOnInit calls navigateToRejected. It could be fixed by using navCtrl.popToRoot instead of setRoot, but then the transition between pages would not be consistent

noro5: I must admit that I do not understand what do you mean. Maybe we can discuss that on a call when you are in the office?

reg0:
@noro5 please re-review
I don't think we'll be able to think of anything more clever here
in fact, I had to do something similar related with push notifications, after either of those is merged, we have to unify those changes

PR #383, opened 2018-04-25
NT-798 - One button component for all action buttons

eiriko2303:


src/components/action-button/action-button.component.html
```diff
@@ -0,0 +1,19 @@
+<button ion-button class="action-button" 
+[disabled]="disabledButton" 
+[ngClass]="{'green-color': class === 'green',
```
reg0: you could have just use class input in class attribute:

```html
<button class="${classes}" ... />
```
```ts
this.classes = `${this.class} ${this.addExtraClass ? this.extraClass : ''}`
```

src/components/action-button/action-button.component.scss
```diff
@@ -0,0 +1,144 @@
...
+
+    .button-md.activated {
+        background-color: transparent !important;
+    }
+
+    .button-right {
+        float: right;
+    }
+
+    .approve-all-btn {
```
reg0: this should probably be not defined here, it's something specific to some view, you should put it in that view's styles

src/components/action-button/action-button.component.scss
```diff
@@ -0,0 +1,144 @@
...
+        height: 35px;
+        float: right;
+        margin-right: 0.8rem;
+        margin-top: 0.5em;
+        visibility: visible;
+        opacity: 1;
+        transition: visibility 0.8s, opacity 0.8s linear;
+    }
+
+    .side-menu-logout-button {
```
reg0: this is also specific to one view I guess?

src/components/action-button/action-button.component.ts
```diff
@@ -0,0 +1,23 @@
...
+@Component({
+    selector: 'action-button',
+    templateUrl: 'action-button.component.html'
+})
+export class ActionButtonComponent {
+
+    @Input() buttonText: string;
+    @Input() icon: string;
+    @Input() iconClass: string;
+    @Input() class: string;
```
reg0: can ```class``` be an identifier? isn't it restricted keyword?
usually one uses ```clazz``` or ```klass``` in such cases

src/components/action-button/action-button.component.html
eiriko2303: Aha i did not now about that setup, really nice 😀 Gonna change it 😃

reg0: not sure if that's exactly how it should be, but I'd go this path ;)

maybe simply like that: ```<button class="classes" ... />```

src/components/action-button/action-button.component.scss
eiriko2303: Yep, moved it 😀

src/components/action-button/action-button.component.scss
eiriko2303: Moved ;)

src/components/action-button/action-button.component.ts
eiriko2303: Ok, i can change the naming :)

src/pages/approval/approval-pm/approval-pm.html
```diff
@@ -18,7 +18,7 @@ <h5 class="period-text">{{approvalService.getWeekName(period.description)}} - {{
               <div class="total-billable">Hours to approve: {{getPeriodSumOfHoursForUserInProject(project.no, user.no, period.startDate)}} h</div>
               <div class="total-registered">Total registered: {{getTotalSummaryOfHoursForPeriodOnProject(user.no, period.startDate, project.no)}} h</div>
               <div class="approve-all-container">
-                <button ion-button class="rounded-btn rounded-btn-green approve-all-btn" (click)="approveAllEntriesInPeriod(project.no, user.no, period.startDate)">Approve all</button>
+                <action-button [buttonText]="'Approve all'" [clazz]="'green-rounded'" [extraClass]="'approve-all-btn'" [addExtraClass]="true" (actionClick)="approveAllEntriesInPeriod(project.no, user.no, period.startDate)"></action-button>
```
reg0: I think in such cases you could combine those classes together inside ```clazz``` attribute, and use ```extraClass``` only when it is really conditional (here it should always be true)

```[clazz]="'green-rounded approve-all-btn'"```

Comment (eiriko2303) [reply src/pages/approval/approval-pm/approval-pm.html]: Ok :)

File: src/components/action-button/action-button.component.ts [hunk 6]
```diff
@@ -0,0 +1,27 @@
...
+    selector: 'action-button',
+    templateUrl: 'action-button.component.html'
+})
+export class ActionButtonComponent {
+
+    @Input() buttonText: string;
+    @Input() icon: string;
+    @Input() iconClass: string;
+    @Input() clazz: string;
+    @Input() extraClass: string;
```
noro5: Maybe a better idea would be instead of using class names as input, make several inputs describing the component like: color (would be a string), transparency (boolean), rounding (boolena) etc? So as a consumer of that component, I do not have to look on the class names, but just configure the component using properties which are self explanatory.

reg0: yup, that's also an option

hasnik: Do we need separate class and extraClass? Why don't we just use one property to provide a set of class e.g. class="row row-grid" and so on?

reg0: @hasnik extraClass is conditional controlled by addExtraClass input
I think it's acceptable, it's slightly cleaner than building that classes string conditionally on "client side"

eiriko2303: Yes the thing was that i needed a class that should be added if a statement is true, so that is for example on the switching of line manager and project manager

PR #384, opened 2018-04-25
NT-889 - Added possibility to open modal with enter key

eiriko2303:


src/pages/welcome-web/welcome-web.ts
```diff
@@ -227,8 +227,10 @@ export class WelcomeWebPage {
       event.preventDefault();
       this.markRelativeActivity(1);
     } else if (event.keyCode === this.KEY_ENTER) {
-      event.preventDefault();
-      this.markRelativeInput();
+      if(document.activeElement.tagName.toLowerCase() === 'input') {
```
reg0: is that enough? I think search field is also an input

noro5: Actually right now everything can be opened with an Enter, which is a nice feature :) The problems is if we can be sure, that all of that works as expected. If we want to be sure that everything works, then just check class/parent, if we would like to keep that functionality, then OK.

eiriko2303: The meaning of using the tagName was to not need to change the function again if we add or remove class names.

eiriko2303: But shoud i then check something other than the tagname or?

PR #387, opened 2018-04-25
NT-900 - Clear inputfield in modal on focus if zero is the value

eiriko2303:


src/components/time-entry-modal/time-entry-modal.component.ts
```diff
@@ -232,6 +232,7 @@ export class TimeEntryModal {
     removeZero(event) {
         this.currentInputValue = 0;
         this.currentInputValue = parseFloat(event.target.value);
+        event.target.value == "0" ? event.target.value = '' : null;
```
reg0: please change it to normal if statement for readability

eiriko2303: Ok, change it :) and also added logic for putting 0 if field has no value when focusout :)

eiriko2303:
@reg0 Now it should be better i think ;)

reg0:
@eiriko2303 now the problem is the other way around :) if the input was empty, you go into it, and blur, "0" will appear, although it wasn't there before

eiriko2303:
@reg0 But is that actually a problem? Like when you have a blank field it is 0 so it is the same value?

reg0:
@eiriko2303 ok, you're right, sorry
if you are already in the modal, there always have to be some value

PR #390, opened 2018-04-26
NT-936 - Added project info to LM and PM in grid

eiriko2303:


src/components/approval-entries-grid/approval-entries-grid.component.scss
```diff
@@ -16,9 +16,12 @@ approval-entries-grid {
             margin-bottom: 30px;
             overflow-x: hidden !important;
         }
-        &-row ion-col{
-            padding-top: 15px;
-            padding-bottom: 15px;
+        &-row {
+            padding-right: 15em;
```
hasnik: Why did you used ems for padding-right, is rest of the layout font scaling ready?

eiriko2303:
@reg0 The responsiveness on the approval page have never been good, so if we should start to fix up in that stuff i think we need a completely different ticket for it. For example the project manager button in the bottom is also disapearing when you are scalling the window, the same for the hours summary in the header.

reg0:
@eiriko2303 but in this particular case, it broke in this branch, it was OK here: http://worktime-test-pr-391.herokuapp.com/

eiriko2303:
@reg0 It was ok, but it is still not good i think :) : 
adadasdasd

reg0:
yet far better than:

eiriko2303:
@reg0 Ok i am going to take a look at it, but then i would also want to add some additional scaling here to this branch so that for example not the "project manager" button disapears and so on

eiriko2303:
@reg0 Sorry for my mistake with this alignment, i have now fixed up in the columns, so it should be good now 👍 I merged the branch with the responsivness fixed just to now that it dont conflicts with the styling :)

reg0:
looks great now!

PR #392, opened 2018-04-26
Line wrapping in my tasks bug

slomianadam:


reg0:
this it the problem which Adam found :)

PR #395, opened 2018-04-26
NT-937 - Added logic to disable the comment button to the input field until the saving of direct hours is finish

eiriko2303:
But i need some testing from you guys and maybe some other suggestions that might be better 😀


src/components/time-input-web/time-input-web.ts
```diff
@@ -195,4 +206,24 @@ export class TimeInputWebComponent {
...
+    this.inputHasFocus = false;
+  }
+
+  userHaveFilledInHours(inputEl) {
+    if(!this.inputHasFocus && !this.showSavingHoursComponent) {
+      this.disableCommentButton = false;
+    }
+
+    if(this.disableCommentButton) {
+      console.log(inputEl);
```
reg0: please remove console log

and, either we don't need the class field and use only return value of this fucntion for disabling, or else we don't need this method and button should disable based on this.disableCommentButton
please choose one way of doing it (probably first one)
and finally, if method stays, then it could have a better name, like "shouldDisableButtonWhileSaving" or similar

eiriko2303: Have removed consolelog and changed the name of the method, but what do you mean with the class field? :)

reg0: @eiriko2303 I meant that ```this.disableCommentButton``` is probably not necessary at all, it is not being used anywhere directly, only the value is set, but whether the button is disabled it only depends on what the function returns, so you could get rid of that field and just return true or false where necessary

src/components/time-input-web/time-input-web.ts
```diff
@@ -34,20 +34,28 @@ export class TimeInputWebComponent {
...
 
   constructor(public modalCtrl: ModalController,
     private dateFormatter: DateFormatPipe,
     private timeEntryService: TimeEntryService,
     private cdr: ChangeDetectorRef,
     public formattingService: FormattingService,
     public saveAnimationService: SaveAnimationService) {
-      
+      this.saving$.subscribe(saving => {
+        this.showSavingHoursComponent = saving.showSavingHoursComponent;
```
noro5: Wouldn't it be sufficient just to set here: disableCommentButton to value saving and bind this value in html to disable attribute? We simply do not want the modal to be opened during saving and I am not sure if we need to check those states focusin and focusout.

eiriko2303: Totally agree, this will be way better, thanks! 👍

src/components/time-input-web/time-input-web.ts
eiriko2303: Working on changing the setup different now after talking with lukasz, so this whole function will not be there ;)

src/services/time-entry.service.ts
```diff
@@ -73,9 +73,11 @@ export class TimeEntryService {
                     }
                     activity.activityTimeEntries.set(Id + '', entry);
                     this.activityService.setActivity(projectNo, activityNo, activity, date);
+                    this.ngRedux.dispatch(deleteDisabledCommentButtonID(projectNo + "-" + activityNo + "-" + date.date.toString()));
```
noro5: I am surprised by this implementation :) In my opinion those calls here are wrong, as service does not care about UI at all, it is only changing the data. I would expect that dispatching of deletedDisacledCommentButtonID would be performed in saveCompletedHandler in the time-input-web.ts. You would dispatch this action only once no matter what action would be performed. Let me know if there is a reason why it is done this way.

eiriko2303: Aha ok, my bad, I though it should be here, but yes, gonna change it now :)

src/components/time-input-web/time-input-web.ts
```diff
@@ -36,18 +38,34 @@ export class TimeInputWebComponent {
...
     private dateFormatter: DateFormatPipe,
     private timeEntryService: TimeEntryService,
     private cdr: ChangeDetectorRef,
     public formattingService: FormattingService,
-    public saveAnimationService: SaveAnimationService) {
-      
+    public saveAnimationService: SaveAnimationService,
+    public ngRedux: NgRedux<IAppState>) {
+      this.disableCommentButtons$.subscribe(disabledButtonIds => {
+        this.disableCommentButtons = disabledButtonIds;
```
noro5: You do not use that collection anywhere, only here for finding if the item should be disabled, so it is not needed to hold that value in a variable, you can use directly the disabledButtonIds which you get from the subscription. Besides that, do not use forEach here (it will iterate over the whole collection), but use: this.isDisabled = disabledButtonIds.some(id => id === this.commentButtonId).

PR #400, opened 2018-04-28
Nt 189: direct registration 

noro5:


src/components/checkbox/checkbox.component.html
```diff
@@ -1,5 +1,6 @@
 <label class="checkbox" (click)="stopPropagation($event)">
     <input type="checkbox" [disabled]="disabled" [ngModel]="value" (ngModelChange)="handleValueChange($event)">
-    <span class="checkbox-checkmark"></span>
-    <span class="checkbox-label">{{label}}</span>
+    <span *ngIf="!arrowMode" class="checkbox-checkmark"></span>
+    <span *ngIf="arrowMode" class="checkbox-arrow"></span>
```
eiriko2303: If it is cust the classes that is different, then maybe just use ngClass?

noro5:
Very good points! 
+ Trash removed
+ comma support implemented
+ worktype added
+ changed button

dalsenius:
The add new entry button is not visible on submitted periods. It should always be visible.
We also need to do something after the user presses save so that they understand what has happend. It says in the JIRA ticket that the user should be sent to the period where the entry was registered

noro5:
Great findings Jonas! Both are fixed now and also I have implemented save animation, so it is more intuitive to the user what is happening and also errors are shown.

src/components/direct-registration-modal/direct-registration-modal.component.ts
```diff
@@ -0,0 +1,194 @@
...
+
+    hasDirectRegistrationDataProjectActivity() {
+        return this.directRegistrationData && this.directRegistrationData.projectNo.length && this.directRegistrationData.activityCode.length;
+    }
+
+    clearDirectRegistrationProjectActivity() {
+        this.directRegistrationService.setProjectActivityForDirectRegistrationData(null, null);
+    }
+
+    showSearchEngingeResult() {
```
reg0: typo

PR #401, opened 2018-04-30
NT-944: fix problem with add another entry and putting hours on all days on g…

noro5:
…iven activity


eiriko2303:
I have made a seperate ticket for updating summay fix, so i am going to merge this now as it fix the largest bug for the release. :)

PR #405, opened 2018-05-03
NT-960 - Added functionality to get rejecters full name for rejected entries

eiriko2303:


src/components/activity-list/activity-single-entry.component.ts
```diff
@@ -133,6 +151,14 @@ export class ActivitySingleEntryComponent {
       return description;
     }
 
+    getRejectersName(act: ActivityModel) {
+        let rejecter = '';
+        act.activityTimeEntries.forEach((val, key, map) => {
```
reg0: doing forEach loop where you know it's only one element is kind of misleading ;)

src/components/time-entry-modal/time-entry-modal.component.ts
```diff
@@ -133,6 +133,9 @@ export class TimeEntryModal {
         if (projectActivityDateEntry && projectActivityDateEntry.activityTimeEntryModels.length > 0) {
             projectActivityDateEntry.activityTimeEntryModels.forEach((atem, key, map) => {
                 let timeEntry = new TimeEntry(atem);
+                this.timeEntryService.getNameOfApprover(timeEntry.notApprovedBy).subscribe(name => {
```
reg0: shouldn't it be filtered out first to make that query only for actually not approved entries?

src/components/activity-list/activity-single-entry.component.ts
eiriko2303: haha, correct, gonna change that 👍

src/components/time-entry-modal/time-entry-modal.component.ts
eiriko2303: Yeah, thats true, maybe i could then just add a if to check if it is a not approved one? :)

eiriko2303: Added it now :)

src/components/activity-list/activity-single-entry.component.ts
```diff
@@ -152,11 +152,7 @@ export class ActivitySingleEntryComponent {
     }
 
     getRejectersName(act: ActivityModel) {
-        let rejecter = '';
-        act.activityTimeEntries.forEach((val, key, map) => {
-          rejecter = val.notApprovedByFullName;
-        });
-        return rejecter;
+        return act.activityTimeEntries[0].notApprovedByFullName;
```
reg0: it's a map, can you take it like that?
maybe:
```Array.from(act.activityTimeEntries.values())[0].notApprovedByFullName;``` ?

eiriko2303: Nice, works fine 👍

PR #407, opened 2018-05-04
NT-920 - Removing whitespace before and after seach query

eiriko2303:


eiriko2303:
@reg0 Yep thats true, going to try fixing that too in this branch, hopefully it would be similare with just using trim :)

eiriko2303:
@reg0 It is added now to every searchfield :)

PR #409, opened 2018-05-07
NT-909 - Showing of approval dont happend if you switch between users at same session

eiriko2303:
- Fix showing of approval at login and also fixed a bug that the approval popup was not showing for manager that dont have the ismanager flag set to true, so now the approval just check if the entriestoapprove.length to now if the approval popup should be shown. The bug with the ismanager flag on the user model is similare to the problem we had earlier that the approval page was not shown for some user, for example the user "maoberg".


hasnik:
👍

PR #410, opened 2018-05-07
NT-919 - Fix label not floating at login

eiriko2303:


reg0:
shrinking animiation is still there when you focus prefilled fields, can we get rid of that somehow?

eiriko2303:
@reg0 Ah, yes I see that too, I can try to fix it, but it was not easy, this is a bit hacky. But I saw that it was also others that had this problems with the floating fields, gonna try :)

eiriko2303:
@reg0 I think it should be good now :)

PR #414, opened 2018-05-09
NT-913 - Add activitycode to places where activitydescription is displayed

eiriko2303:
I think i have added it to the places where it should be, activity list, search, approval modul, modal.


noro5:
Works very nicely, but I found small issues:
- I think it is missing in the Favourite List (left side in the Task Manager)
- small improvement in alignment in the Activity List in mobile would be needed:
image

eiriko2303:
@noro5 Great found, fixed it now :)

PR #416, opened 2018-05-11
NT-963 - No supported browser page

eiriko2303:


src/pages/not-supported-browser/not-supported-browser.scss
```diff
@@ -0,0 +1,26 @@
+not-supported-browser {
+    .nosupport {
+        &-content, &-footer {
+            &-image, &-text {
+                display: block;
+                margin-left: auto;
+                margin-right: auto;
```
hasnik: I believe _margin: auto;_ will be sufficient

hasnik: This site actually looks awful on mobile

src/app/app.component.ts
```diff
@@ -16,6 +16,21 @@ import { setHasShownDownloadAlert } from '../actions/rootActions';
 import { AppVersion } from '@ionic-native/app-version';
 import { OneSignal } from '@ionic-native/onesignal';
 
+//Values that is browser specific
```
hasnik: I believe the code for no supported browser should be enclosed in some other class, not in app.component

src/pages/not-supported-browser/not-supported-browser.scss
eiriko2303: ok :)

src/app/app.component.ts
eiriko2303: How should it be used? It needs to be called first i think?

eiriko2303: Ok i have placed the declaring of the variables in a seperate class, but the method needs to stay in app.component cause it needs to be available imediately at start of the app.

hasnik:
I appreciate the changes, but as Kamil said. 'isSupportedBrowser' function should not belong to app.component

eiriko2303:
Yes i understand, but the problem i got is how should i then have the function as a entry function when the app is loaded?

reg0:
I fixed it 
@hasnik please re-review

eiriko2303:
@reg0 Great work! thanks 😄

hasnik:
@reg0 @eiriko2303 nice work!

PR #417, opened 2018-05-15
NT-967 - Added possibility to search for activityCode

eiriko2303:


reg0:
@eiriko2303 shouldn't it also be included in web? (activity.service.ts -> matchesSearchTerm())

eiriko2303:
@reg0 Oh, sorry did not see that it was also one there to, thanks! 👍

PR #422, opened 2018-05-16
NT-973 - New styling of timeentries in activitylist and also some minor css styling to the top-toolbar

eiriko2303:


eiriko2303:
Great that you saw that! I did not test with worktype dropdown, gonna fix it :)

PR #423, opened 2018-05-18
NT-974 - Side menu icons and 'W' instead of full Worktime logo on mimimizing of side menu

eiriko2303:


eiriko2303:
The problem with the centering of the icons was the notification icon about hours to approve so i just made it to fit inside the side menu, but maybe we can just not have it when the menu is minimized? Yeah i did think of that too, i have sent Bodil the link to the page to get her feedbacks on it :)

eiriko2303:
@reg0 I got the logout icon from Bodil now so i have added it, and also centered the other icons after removing the notification icon when minimizing the menu. Bodil said that the notification icon is not important when you minimize, but should stay when it is open.

reg0:
Looks much better now :)

PR #427, opened 2018-05-23
NT-982 - Removing white page after splash screen startup animation

eiriko2303:
- Have added values to config.xml to have the splash screen stayed until the app has completely load and also small fade in animation for nicer user experience


eiriko2303:
@reg0 Yeah, so this changes is for the splashscreen to be be shown until it is beeing hided in platform.ready (app.component), so then the app will have finish the loading of the content before we hide the splashscreen :)

PR #428, opened 2018-05-23
NT-980 - Dropdown to switching between approval pages on mobile

eiriko2303:
This is a branch that is created out of base branch NT-976: "NT-976 - Approval mobile order setup for PM and LM".


reg0:
same as NT-978 -> I'd merge to 979 and close this PR

PR #430, opened 2018-05-28
NT-985 - Styling to settings pages and added back to settings button

eiriko2303:


src/pages/my-devices/my-devices.ts
```diff
@@ -56,4 +57,7 @@ export class MyDevicesPage {
     });
   }
 
+  goToSettings() {
+    this.navCtrl.push(SettingsPage);
```
reg0: couldn't we somehow pop here, instead of pushing next instance of SettingsPage?
currently for each click on settings page, one of subpages, back to settings, and so on, you get multiplied html

eiriko2303: Oh, i did not see that! Great find, and i fixed it now, it should just pop back! 👍

PR #431, opened 2018-05-29
#NT-892: Add keyboard shortcuts

hasnik:


src/components/comment-toggle/comment-toggle.ts
```diff
@@ -19,6 +20,7 @@ export class CommentToggleComponent implements OnInit {
   selectedPeriodIndex$: Observable<number>;
   selectedPeriodIndex: number;
 
+  private readonly KEY_C = 67;
```
reg0: I would name those constants with something meaningful (and maybe add a comment about actual letter), so that in case we want to change the letter, we just change the value here and not all it's occurrences:

```ts
private readonly KEY_CODE_COMMENTS_TOGGLE = 67; // 'C'
```

src/components/comment-toggle/comment-toggle.ts
```diff
@@ -45,4 +47,11 @@ export class CommentToggleComponent implements OnInit {
     return true;
   }
 
+  @HostListener('document:keydown', ['$event'])
+  public keyPressed(event: KeyboardEvent) {
+    if (event.altKey && event.keyCode === this.KEY_C) {
+      event.preventDefault();
```
reg0: wouldn't this disallow to write "c" character anywhere? it looks like a global event listener on a document, so if you preventDefault(), "c" will not be typed in search or comment field

src/components/logout-button/logout-button.ts
```diff
@@ -22,6 +22,8 @@ export class LogoutButtonComponent implements OnInit {
   private user$: Observable<UserModel>;
   user: UserModel;
 
+  private readonly KEY_L = 76;
```
reg0: same as above, let's name the constants according to their purpose

reg0: and also, as we talked yesterday, maybe alt + X or alt + Q would be a better hotkey for logout?

src/components/submit-reopen-period/submit-reopen-period.component.ts
```diff
@@ -66,4 +68,28 @@ export class SubmitReopenPeriod {
     getButtonText() {
         return "Submit week " + this.periodService.getWeekNumber(this.selectedPeriodIndex);
     }
+
+    hideSubmitButton() {
```
reg0: these 3 methods could have better names, getter-like ones:

isReopenButtonHidden, isSubmitButtonHidden, isComponentVisible 

with that, their usage would look more natural and readable

src/components/comment-toggle/comment-toggle.ts
noro5: I think we are good here, as there is: _event.altKey &&_

eiriko2303: Yeah I agree with Lukasz, i think it is ok, because it will just be true if c is hit at the same time as also alt is hit.

reg0: alt+c and alt+l are actually key combinations in Polish keyboard layout to produce our special characters "ć" and "ł" :)
but OK, I can assume that even if we're using Worktime, we won't write comments in Polish in the app :D

reg0:
@eiriko2303 I don't even try to imagine clean implementation of this :)
of course, it'd be nice, but in our case, all these actions refer to something else and buttons are not grouped together like in the screenshot you posted

maybe another way, would be more oldschool - underline below the letter on appropriate buttons _L_ogout, _S_ubmit, etc.

src/components/comment-toggle/comment-toggle.ts
eiriko2303: Oh, that i did not know, but maybe we can have a check if it is inside the commentfield? So that when you are inside the commentfield the shortcuts would not work? :)

eiriko2303:
@reg0 Yeah it was just an example, i think that maybe we can have something similare to just showing the different shortcuts so that the user know about them.
asd

PR #432, opened 2018-05-30
NT-1006 - Side menu larger click area to toggle.

eiriko2303:


src/app/app.component.ts
```diff
@@ -321,4 +324,14 @@ export class MyApp {
     }
 
   }
+
+  toggleSideMenu() {
+    let element = document.getElementsByClassName("worktime-menu")[0];
+    this.sideMenuHidden = !this.sideMenuHidden;
+    if (this.sideMenuHidden) {
+      this.renderer.addClass(element, "collapsed");
```
reg0: why didn't you just add that class in ngClass in app.html?

eiriko2303: Dont know why I did that, but yes it is much cleaner and simplare with the ngClass 😑 Great find, thanks! 😄 👍

reg0: @eiriko2303 so now I think passing and injecting Renderer is not needed anymore?

eiriko2303: 👍

src/components/page-list/page-list.ts
```diff
@@ -143,7 +143,8 @@ export class PageListComponent implements OnInit {
     return this.forWeb && this.isApprover;
   }
 
-  openPage(page) {
+  openPage(page, event?) {
+    event ? event.stopPropagation() : null;
```
reg0: @eiriko2303 
please use that in ```keyPressed``` -> pass event from there to ```openPage``` instead of calling ```stopPropagation()``` in each if statement

reg0: ah, sorry, I didn't notice the difference
there's preventDefault and here's stopPropagation

eiriko2303: Yep, so the stopPropagation is to not having the user closing the menu when they want to go to another page :)

PR #434, opened 2018-05-31
NT-962 - Get entries that line manager should not approve

eiriko2303:


src/pages/approval/approval-lm/approval-lm.ts
```diff
@@ -187,4 +203,38 @@ export class ApprovalLmComponent implements OnInit{
...
+  toggleOtherEntries(resourceNo: string, startDate: Date) {
+    if(this.toggleOtherEntriesOpenID === this.getIdentifierForOtherEntries(resourceNo, startDate)) {
+      this.toggleOtherEntriesOpenID = "";
+    } else {
+      this.toggleOtherEntriesOpenID = this.getIdentifierForOtherEntries(resourceNo, startDate);
+    }
+  }
+
+  isOtherEntriesOpen(resourceNo: string, startDate: Date) {
+    return resourceNo + "-" + startDate.toISOString() === this.toggleOtherEntriesOpenID ? true : false;
```
reg0: @eiriko2303 
here you could also use ```getIdentifierForOtherEntries()```
and also, you don't need ternary operator if you just do ```condition ? true : false```, you should just return result of comparison

eiriko2303: Ah, off course, my mistake! Was a bit guick there, fixed it now 👍

eiriko2303:
@reg0 Yep that was my idea to, but I wanted to ask Kenneth today during the standup cause i dont know if we should hide the table or not, cause maybe the line managers then think there is something wrong. But probably it would be best to just hide it.

eiriko2303:
@reg0 The other entries table is replaced with an illustration with text when it is empty. Should be ok now i think :)

PR #436, opened 2018-06-08
NT-376 - Android physical back button compability

eiriko2303:
Back button can do the following:
- Go back from pages
- Close side menu
- Close dropdown
- Close the customkeyboard


src/app/app.component.ts
```diff
@@ -47,97 +53,150 @@ export class MyApp {
...
+      } else {
+        this.checkWhichPageToPop();
+      }
+    })
+  }
+
+  checkWhichPageToPop() {
+    let max = pages.length;
+    for (let i = 0; i < max; i++) {
+      let n = this.navCtrl;
```
noro5: Maybe that definition should be outside the For loop?

src/components/week-picker-dropdown/week-picker-dropdown.component.html
```diff
@@ -1,4 +1,4 @@
-<div class="dropdown-list" *ngIf="showDropdown">
+<div class="dropdown-list" *ngIf="showDropDown">
```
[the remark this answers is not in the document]
eiriko2303: I have changed the naming to camelcase cause i thought it should be like that, but I can change it back? :)

src/app/app.component.ts
eiriko2303: Agree, fixing it now :)

src/app/app.component.ts
```diff
@@ -47,97 +53,150 @@ export class MyApp {
...
+        return;
+      } else if(this.showingKeyboard) {
+        this.ngRedux.dispatch(setShowKeyboard(false));
+      } else {
+        this.checkWhichPageToPop();
+      }
+    })
+  }
+
+  checkWhichPageToPop() {
```
reg0: I would name it differently, "check" suggests it returns boolean, and doesn't do any other action, but that's not the case

how about ```popPageIfPossible()``` ?

src/components/week-picker/week-picker.ts
```diff
@@ -124,15 +127,15 @@ export class WeekPickerComponent {
   }
 
   showDropDown() {
-    if(this.weekPickerDropdown.showDropdown) {
-      this.weekPickerDropdown.showDropdown = false;
+    if(this.isDropDownShown) {
```
reg0: method should be renamed to toggleDropdown, if it also hides it

src/app/app.component.ts
eiriko2303: Agree :)

src/components/week-picker/week-picker.ts
eiriko2303: 👍

src/app/app.module.ts
```diff
@@ -96,6 +96,8 @@ import { ActionButtonComponent } from '../components/action-button/action-button
 import { ApprovalHistoryPage } from '../pages/approval-history/approval-history';
 import { BrowserSupport } from './browserSupport';
 import { DirectRegistrationModal } from '../components/direct-registration-modal/direct-registration-modal.component';
+import { InitalizationService } from '../services/initialization.service';
```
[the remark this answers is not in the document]
eiriko2303: Thanks! Fixed and tested it now :)

PR #437, opened 2018-06-11
Nt 44 modal

noro5:


src/components/copy-last-day-modal/copy-last-day-modal.component.ts
```diff
@@ -0,0 +1,118 @@
...
+        this.worktypes$.subscribe(data => this.worktypes = data);
+        this.worktypeMode$.subscribe(data => this.showWorktype = data);
+    }
+
+    showKeyboard(event) {
+        this.customKeyboard.showKeyboard(event);
+    }
+
+    keyboardDone(amountOfHours: number) {
+    }
```
eiriko2303: What is this empty function for? :)

src/pages/welcome/welcome.html
```diff
@@ -7,6 +7,7 @@
     [currentPeriodIndex]="currentPeriodIndex" (filterArrayChanged)="onFilterArrayChanged($event)"></top-menu-mobile>
 </ion-header>
 <ion-content *ngIf="periods && selecteddate && getSelectedPeriod()" id="welcome-content" #content>
+  <button *ngIf="getSelectedPeriod().isOpen()" [disabled]="!hasTimeEntriesOnPreviousDay()" class="copy-last-day"ion-button (click)="copyLastDay()">Copy Last Day</button>
```
eiriko2303: Should maybe be the action button component, but we will see what Bodil comes up with :)

src/pages/welcome/welcome.ts
```diff
@@ -553,4 +555,60 @@ export class WelcomePage implements OnInit {
   handlePeriodChanged() {
     this.updateSlidesLocks();
   }
+
+  copyLastDay() {
```
eiriko2303: I think that the stuff for copy last day should be in a seperate component. The welcome.ts file is quite large already.

src/pages/welcome/welcome.html
eiriko2303: And this button could also be placed in a seperate component.

eiriko2303:
Possible bug in the styling of the modal in the button section in the bottom: 
dfsdfsdf

src/components/copy-last-day-modal/copy-last-day-modal.component.ts
reg0: why is it empty? or else, why do we need if it's empty? :)

reg0:
I see that @eiriko2303 already described everything that I also wanted to write :D
and as discussed, pity that we couldn't extract some common part for displaying single entry's details between this and time-entry-modal, they look basically the same, and ideally would have common html+css

src/pages/welcome/welcome.html
noro5: Yes, let's wait for Bodil - I would use here _Action-Button_ component.

src/pages/welcome/welcome.ts
noro5: I disagree on that - this function is simply a click handler for a button, so what would be the purpose of that component? Maybe we can use _Action-button_ component for the button (depends on what Bodil will create), but this function is handling open/dismiss modal, so I think there is no point to have a component for that. What is more, we have here access to reloadList - with a component we would have to emit an event to call that...

eiriko2303: What i meant was not just the copyLastDay() method, but also the functions hasTimeEntriesOnPreviousDay(), getDateModelForYesterday() and getTimeEntriesFromPreviousDay(), so that we would have less code in the page.

noro5: Yep, I will have a look on them, maybe it will be good to have it in a service dedicated for Copy Last Day.

noro5:
All problems should be solved now :)

src/components/copy-last-day-modal/copy-last-day-modal.component.ts
eiriko2303: Again, does we need to have this empty function here? :)

src/pages/welcome/welcome.html
eiriko2303: Yep :)

src/pages/welcome/welcome.ts
```diff
@@ -553,4 +557,36 @@ export class WelcomePage implements OnInit {
...
+    copyLastDayModal.present();
+    copyLastDayModal.onDidDismiss(data => {
+      if (data.shouldReload) {
+        this.reloadList();
+      }
+    })
+  }
+
+  hasTimeEntriesOnPreviousDay() {
+    const dateModelForYesterday = this.copyLastDayService.getDateModelByDate(this.getYesterdayDate());
```
eiriko2303: Great to have it in a service! 💃

src/pages/welcome/welcome.ts
```diff
@@ -553,4 +557,36 @@ export class WelcomePage implements OnInit {
...
+
+    return false;
+  }
+
+  private getYesterdayDate() {
+    return moment(this.selecteddate).add(-1, 'days').toDate();
+  }
+
+  private getTimeEntriesFromPreviousDay(yesterday: Date): Array<BaseEntry> {
+    return this.copyLastDayService.getTimeEntriesByDate(yesterday);
```
eiriko2303: Nice and clean 👍

PR #440, opened 2018-06-13
NT-978 - Approval entries for PM and LM on mobile

eiriko2303:


reg0:
I'd say this PR should be closed to avoid confusion. You have already merged those branches in all directions and they contain mostly the same changes, so there is no point in maintaining and reviewing them separately, they can't be extracted back at this point anymore

I'd say, NT-979 should be the main one, all those other things related with approvals should be merged there and finally merged to dev when all is ready

eiriko2303:
Ok, i will merge all changes to it and then close the other ones :)

PR #444, opened 2018-06-21
NT-76 - Statistic page

mariusch:
As of 03.07 this includes: 
NT-1024 - The PBR percentage
NT-1027  - Statistics toolbar (Green bar, dropdown for changing periods and dynamic text)
NT-1030 - Toggle for statistics toolbar (later changed in NT-76)
NT-1029 - Buttons for PBR and Away from work pages
NT-1031 - Table for periods (NT-1048 - Horizontal bars is NYI)
NT-1032 - Stacked bar chart


src/components/period-toggle/period-toggle.ts
```diff
@@ -0,0 +1,38 @@
+import { Component, Output, Input, EventEmitter } from '@angular/core';
+
+@Component({
+  selector: 'period-toggle',
+  templateUrl: 'period-toggle.html'
+})
+export class PeriodToggleComponent {
```
reg0: I would name it YearMonthToggle because PeriodToggle can be very misleading given how important word "period" is in this project

src/components/period-toggle/period-toggle.ts
```diff
@@ -0,0 +1,38 @@
...
+  @Input() periodType: boolean;
+  @Output() periodChange: EventEmitter<boolean> = new EventEmitter<boolean>();
+
+  periodTypeLocal: string; //currently we need this because ion-button only works with strings
```
reg0: why do we need this local value here? it doesn't seem necessary
value in the html could be set based on the boolean (or enum) value, [(ngModel)] also doesn't seem to improve a lot here if there is separate changeHandler

src/components/period-toggle/period-toggle.ts
```diff
@@ -0,0 +1,38 @@
+import { Component, Output, Input, EventEmitter } from '@angular/core';
+
+@Component({
+  selector: 'period-toggle',
+  templateUrl: 'period-toggle.html'
+})
+export class PeriodToggleComponent {
+  @Input() periodType: boolean;
```
reg0: maybe it'd be more readable if there was some special enum here? boolean for year/month is not self explanatory

src/components/statistics-table/statistics-table.html
```diff
@@ -0,0 +1,28 @@
...
+        <ion-row class="activity-grid-title">
+            <ion-col class="entries-col">
+                <span class="dot" [style.background]="circleColor"></span>
+                <div class="title">{{ title + " (" + getTotalHours() +")" }}</div>
+            </ion-col>
+        </ion-row>
+        <ion-row class="activity-grid-list">
+            <ion-toolbar id="activity-grid-head" class="activity-grid-header-top">
+                <ion-row class="activity-grid-header labels" nowrap>
+                    <ion-col class="entries-col">KUNDENAVN</ion-col>
```
reg0: all labels should be in English

package.json
```diff
@@ -40,7 +40,10 @@
     "@ionic-native/splash-screen": "^4.7.0",
     "@ionic-native/status-bar": "^4.7.0",
     "@ionic/storage": "^2.1.3",
+    "angular2-chartjs": "^0.5.1",
```
eiriko2303: Remove unused library

src/components/statistics-toolbar/statistics-toolbar.ts
```diff
@@ -0,0 +1,190 @@
...
+      }
+    })
+  }
+  
+  setArrayOfMonths() {
+    let iterateMonth = 0;
+    this.monthArray = [];
+
+    this.periods.forEach((period) => {
+      if (iterateMonth !== period.startDate.getMonth()) {
```
reg0: is this enough if periods are across 2 years?

package.json
eiriko2303: Remove unused library

src/components/bar-diagram/bar-diagram.component.ts
```diff
@@ -0,0 +1,244 @@
...
+  nonBillableEntriesForStatistics: Array<EntryForStatistics>;
+
+  @Input() statTables: Array<any>;
+  @Input() normTimeForPeriod: number = 0;
+
+  totalAllHours: number = 0;
+  totalDeviationHours: number = 0;
+  totalBillabeEntries: number = 0;
+  totalNonBillabeEntries: number = 0;
+  totalAwayEntries: number = 103; //Missing
```
eiriko2303: Set default value to be 0, although we dont have this data yet.

src/components/bar-diagram/bar-diagram.component.ts
```diff
@@ -0,0 +1,244 @@
...
+    entries.forEach(entry => {
+      count += entry.quantity;
+    });
+    return count;
+  }
+
+  initDataForChart() {
+    this.context =  (<HTMLCanvasElement>this.stackedBar.nativeElement).getContext('2d');
+
+    var gradient = this.context.createLinearGradient(0, 0, 0, 100);
```
eiriko2303: Maybe remove the unused gradient or try to find a solution where we can use it to show stripes instead of color :)

src/components/statistics-toolbar/statistics-toolbar.ts
```diff
@@ -0,0 +1,190 @@
...
+  getPeriodRangeArray() {
+    if(this.periodType === false) {
+      return this.yearArray;
+    } else {
+      return this.monthArray;
+    }
+  }
+
+  getLastDayOfMonth(year, month) {
+    return new Date(year, month+1, 0).getDate();
```
reg0: this doesn't seem right, looks confusing, and will probably break due to timezones
please use moment.js for all such dates manipulating stuff

src/components/details-toggle/details-toggle.ts
```diff
@@ -0,0 +1,25 @@
+import { Component, Input, Output, EventEmitter, APP_INITIALIZER, SimpleChanges } from '@angular/core';
```
eiriko2303: Remove APP_INITIALIZER from import. It is not used?

src/components/statistics-table/statistics-table.ts
```diff
@@ -0,0 +1,34 @@
...
+    let value = "";
+    value += `${CircleColor[this.circleColor].toLowerCase()}`;
+    return value;
+  }
+
+  getGroupedSortedEntries() {
+    return this.statService.getGroupedSortedEntries(this.pipedEntries);
+  }
+
+  getTotalHours() {
```
eiriko2303: This could be moved to the service so that it can be used from this component and also the bar chart to get the total amount of quantity.

src/components/statistics-toolbar/statistics-toolbar.ts
```diff
@@ -0,0 +1,190 @@
...
+    return diffDays > 31;
+  }
+
+  isSelectedPeriodAllYears(firstDate: Date, lastDate: Date): boolean {
+    var timeDiff = Math.abs( lastDate.getTime() - firstDate.getTime());
+    var diffDays = Math.ceil(timeDiff / (1000 * 3600 * 24)); 
+    return diffDays > 366;
+  }
+
+  getSuffix(i: number): string {
```
reg0: this could be put in formattingService and names getOrdinalSuffix

src/components/statistics-toolbar/statistics-toolbar.ts
```diff
@@ -0,0 +1,190 @@
...
+    return menuTitle;
+  }
+
+  isSelectedPeriodAllMonths(firstDate: Date, lastDate: Date): boolean {
+    var timeDiff = Math.abs( lastDate.getTime() - firstDate.getTime());
+    var diffDays = Math.ceil(timeDiff / (1000 * 3600 * 24)); 
+    return diffDays > 31;
+  }
+
+  isSelectedPeriodAllYears(firstDate: Date, lastDate: Date): boolean {
```
reg0: it will probably not work in employee's first year of work - his all years will still be one year and most likely less than 366 days

src/components/statistics-toolbar/statistics-toolbar.ts
```diff
@@ -0,0 +1,190 @@
...
+        this.monthArray.push(monthObject);
+        iterateMonth = period.startDate.getMonth();
+      }
+    })
+    //check if last period ends in a different month than it starts in. If so, push one more month to the array.
+    let lastPeriod = this.periods[this.periods.length - 1];
+    if (lastPeriod.endDate.getMonth() !== lastPeriod.startDate.getMonth()) {
+      let monthObject = new PeriodForStatistics(
+        new Date(lastPeriod.endDate.getFullYear(), lastPeriod.endDate.getMonth(), 1),
+        new Date(lastPeriod.endDate.getFullYear(), lastPeriod.endDate.getMonth(), lastPeriod.endDate.getDate())
```
reg0: why couldn't you just clone ```lastPeriod.endDate``` ?
with that, you introduced possible problem with timezones (local midnight is not always proper midnight)

src/pages/statistics/statistics.ts
```diff
@@ -0,0 +1,105 @@
...
+  }
+
+  ngOnInit() {
+    if (this.isWeb) {
+      this.viewCtrl.showBackButton(false);
+    }
+}
+
+  ionViewDidEnter() {
+    let startDate = new Date(2018,1,1);
```
eiriko2303: The data is for the full year (2018), but the period toolbar is set to be the current month (july). Set instead the initalize data to be current month cause that is what the toolbar is showing. Maybe just call the initalizestate from the service? :)

src/models/periodForStatistics.model.ts
```diff
@@ -0,0 +1,16 @@
+export class PeriodForStatistics {
```
reg0: I wouldn't call it a Period, it's somehow reserved word in this project an has a special meaning
how about DatesRangeForStatistics?

src/models/periodForStatistics.model.ts
```diff
@@ -0,0 +1,16 @@
...
+
+    private monthNames = ["January", "February", "March", "April", "May", "June",
+    "July", "August", "September", "October", "November", "December"
+    ];
+    
+    constructor(public firstDate: Date, public lastDate: Date) {}
+
+    getMonthOfFirstDate(): string {
```
reg0: maybe ```getMonthNameOfFirstDate``` ?

src/models/periodForStatistics.model.ts
```diff
@@ -0,0 +1,16 @@
+export class PeriodForStatistics {
+
+    private monthNames = ["January", "February", "March", "April", "May", "June",
```
reg0: month names formatting should be done using moment.js

eiriko2303:
@mariusch @sebdak  Added some comments for improving the functionality, but other than that the functionality works great I think. So really good job guys! 😄 🥇 👍

src/pages/statistics/statistics.ts
reg0: why is it hardcoded ?
timezones not taken into consideration, possible midnight problem

src/services/statistics.service.ts
```diff
@@ -0,0 +1,116 @@
...
+            return 1;
+            }
+            return 0;
+        })
+    }
+
+    getInitialPeriod() {
+        let currentDate = new Date();
+        return new PeriodForStatistics(
+            new Date(currentDate.getFullYear(),currentDate.getMonth(),1),
```
reg0: timezones, midnights

src/services/statistics.service.ts
```diff
@@ -0,0 +1,116 @@
...
+
+    getNormTime(): string {
+        return  this.normTimeForPeriod + "";
+    }
+
+    getTotalVacationHours(): string {
+        let sum = 0;
+        if(this.nonBillableEntriesForStatistics) {
+            this.nonBillableEntriesForStatistics.forEach(entry => {
+                if (entry.projectNo === "01") {
```
reg0: such values shouldn't be hardcoded, it should be put in the configuration (maybe in environment.ts)

src/pages/statistics/statistics.ts
```diff
@@ -0,0 +1,105 @@
...
+    let endDate = new Date(2018,11,31);
+    this.getEntriesForPeriod(startDate, endDate);
+  }
+
+  getEntriesForPeriod(startDate : Date, endDate : Date) {
+    this.statService.getEntriesForStatistics(startDate, endDate).subscribe(res => {
```
reg0: res and res2 are not meaningful here, especially when they are in nested iterations, please name them cleaner

src/components/bar-diagram/bar-diagram.component.ts
```diff
@@ -0,0 +1,244 @@
...
+      //Hidden bar
+      this.data.datasets[4].data[1] = this.normTimeForPeriod.toFixed(2);
+      this.data.datasets[5].data[1] = Math.abs(this.totalDeviationHours < 0 ? this.totalDeviationHours : 0);
+      this.data.labels = [this.normTimeForPeriod, this.totalAllHours];
+      
+      this.chart.update();
+    }
+  }
+
+  getTotalQuantityOfEntries(entries: Array<EntryForStatistics>): number {
```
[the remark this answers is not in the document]
reg0: if not, please create it and reuse where possible
those classes all extend BaseEntry which contains quantity field, no need to create separate implementation for all subclasses

src/actions/rootActions.ts
```diff
@@ -204,3 +205,16 @@ export function deleteDisabledCommentButtonID(disableCommentButtonId: string) {
 export function setWeekPickerDropDownIsShowing(weekPickerDropDownIsShowing: boolean) {
   return { type: reducerStuff.SET_WEEK_PICKER_DROP_DOWN_IS_SHOWING, weekPickerDropDownIsShowing };
 }
+
```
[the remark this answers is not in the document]
mariusch: Got any comments on this @eiriko2303 ?

src/components/bar-diagram/bar-diagram.component.ts
```diff
@@ -0,0 +1,244 @@
...
+import { Observable } from "rxjs";
+
+@Component({
+  selector: "bar-diagram",
+  templateUrl: "bar-diagram.component.html"
+})
+export class BarDiagram {
+  @ViewChild('stackedBar') stackedBar: ElementRef;
+  public context: CanvasRenderingContext2D;
+  chart: Chart = [];
```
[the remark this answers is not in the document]
mariusch: Im not really sure, but I think it is needed if updateData() is triggered before initDataForchart(). The page breaks if I change it

src/components/statistics-toolbar/statistics-toolbar.ts
sebdak: I think it should work as long as the duration of one single period is not longer than 11 months. Can add a check on the year also, just to be sure.

src/components/bar-diagram/bar-diagram.component.ts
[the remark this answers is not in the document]
eiriko2303: Yeah the initialization must be done at the moment the chart is beeing used. This is how the default setup for chart.js in Angular looks like. Cause in Angular we need to bind the chart object with the canvas element in the html.

src/actions/rootActions.ts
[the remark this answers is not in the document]
eiriko2303: Hi @mariusch and @reg0 :) I dont know exactly if it should be in redux or in the statistic service. But it would be used in other pages, we have this two links on the statistic page that is going to send the user to the "personal billing rate page" and the one for "away from work page". The arrays with filtered entries would be used in both this pages. The most important part of why we added it to redux was because of the initialization of the bar chart. We need to know when the arrays is available and then run the function for adding the data into chart.js and update it.

src/pages/statistics/statistics.ts
```diff
@@ -0,0 +1,107 @@
...
+  getEntriesForDatesRange(startDate : Date, endDate : Date) {
+    this.statService.getEntriesForStatistics(startDate, endDate).subscribe(entries => {
+      this.normTimeForPeriod = this.statService.normTimeForPeriod;
+      this.statService.getInternalChargableTasks().subscribe(internalChargableTasks => {
+        entries.forEach(entry => {
+          if(internalChargableTasks.some((internalProject) => { return entry.projectNo == internalProject.ProjectNo; })) {
+            entry.chargeable = true;
+          }
+        });
+        this.ngRedux.dispatch(setBillableEntriesForStatistics(entries.filter(entry => entry.chargeable === true)));
```
noro5: It seems that it is not needed to hold these values in Redux, because they are used only here and in the service, but in the service they are not needed.

src/components/statistics-table/statistics-table.ts
mariusch: This is already done in the NT-1048 branch :)

src/components/bar-diagram/bar-diagram.component.ts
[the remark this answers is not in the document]
mariusch: We've made improvements to better reuse existing functions in the NT-1048 branch, such as in this case

src/components/statistics-toolbar/statistics-toolbar.ts
mariusch: Removed this altogether and used moment.js

src/services/statistics.service.ts
mariusch: Moved to environment.ts

src/components/bar-diagram/bar-diagram.component.ts
```diff
@@ -0,0 +1,241 @@
...
+      this.updateData();
+    });
+    this.nonBillableEntriesForStatistics$.subscribe(statNonBillableEntries => {
+      this.nonBillableEntriesForStatistics = statNonBillableEntries;
+      this.totalNonBillableEntries = this.getTotalQuantityOfEntries(this.nonBillableEntriesForStatistics);
+      this.updateData();
+    });
+  }
+
+  ngOnChanges() {
```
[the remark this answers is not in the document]
mariusch: There is some problems with normTime not updating correctly if is's done only in the constructor. I made some changes to better match how it is done in activity-list.component.ts

src/services/statistics.service.ts
```diff
@@ -0,0 +1,106 @@
...
+    }
+
+   calculatePBRForPeriods(): string {
+        let billableHours = this.getTotalBillableHours();
+        let normTime = this.getNormTime();
+        let vacationHours = this.getTotalVacationHours();
+        return this.formattingService.formatOneDecimals(parseFloat(billableHours)/(parseFloat(normTime)-parseFloat(vacationHours))*100);
+    }
+
+    getTotalBillableHours(): string {
```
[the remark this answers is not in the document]
mariusch: This is already done in NT-1048 :)

src/pages/statistics/statistics.ts
mariusch: As discussed on the standup today we're puting this on hold until we see how the other pages will be

src/components/statistics-toolbar/statistics-toolbar.ts
mariusch: You're right, I rewrote this

src/components/statistics-toolbar/statistics-toolbar.ts
mariusch: Changes this to momentjs. Let us know if this should be done differently

src/services/datesService.ts
```diff
@@ -0,0 +1,15 @@
...
+
+@Injectable()
+export class DatesService {
+
+    midnightMoment = (date) => {
+        return moment(date).hours(0).minutes(0).seconds(0).milliseconds(0).add(-date.getTimezoneOffset(), 'minutes');
+     };
+
+     midnightDate = (date) => {
+        return moment(date).hours(0).minutes(0).seconds(0).milliseconds(0).add(-date.getTimezoneOffset(), 'minutes').toDate();
```
reg0: midnightMoment should be reused here

src/services/datesService.ts
```diff
@@ -0,0 +1,15 @@
+import { Injectable } from '@angular/core';
+import * as moment from 'moment';
+
+@Injectable()
+export class DatesService {
+
+    midnightMoment = (date) => {
+        return moment(date).hours(0).minutes(0).seconds(0).milliseconds(0).add(-date.getTimezoneOffset(), 'minutes');
```
reg0: I'm wondering if there shuoldn't be .utc() added somewhere at the end
it would be good to check whether

midnightMoment(new Date(2018, 2, 1)).toDate().getMonth() still returns 2? I'm not sure which way could it fail (in summer or winter?), but I can imagine that converting it to midnight will end up with previous day, 23:00 and then month number can be invalid

this method only makes timestamps correct and database comparisons on whole dates, but getting portions of date with getters might still be wrong :/

mariusch: From my testing  (and undertanding https://stackoverflow.com/a/17856735) both
moment(date).hours(0).minutes(0).seconds(0).milliseconds(0).add(-date.getTimezoneOffset(), 'minutes').utc(); 
and 
return moment(date).hours(0).minutes(0).seconds(0).milliseconds(0).add(-date.getTimezoneOffset(), 'minutes'); 

returns the value 2, as well as the same hour. I find it hard to wrap my head around how the outcome would be with summer and winter time as Im not exactly sure how it is saved (https://stackoverflow.com/a/2532962), but I guess getTimezoneOffset() will take that into account?

reg0:
I found 2 more things 
- when you first load the statistics page, year/month toggle has no value and no sliding marker - this is maybe not directly related with this ticket, but I noticed it now (I saw it in #1070 branch but I guess it will be the same here), and I put it here because it's more related with statistics page, and not pbr page
- please check styles in Firefox, I see that lines in the chart are slightly misaligned

PR #448, opened 2018-06-26
NT-1035: implement setting for CopyLastDay, new styling

noro5:


eiriko2303:
Here is two pictures of the two design issues i mentioned on ios:
ddddd
aaaa

PR #453, opened 2018-06-29
NT-1007 - Add popup for unsubmitted periods

eiriko2303:


src/pages/welcome-web/welcome-web.ts
```diff
@@ -177,7 +180,22 @@ export class WelcomeWebPage {
     });
   }
 
-  navigateToApproval(showPopup: boolean, rejectedNotFoundCallback?: Function) {
+  navigateToFirstUnsubmittedPeriod(showPopup: boolean, rejectedNotFoundCallback?: Function) {
```
noro5: rejectedNotFundCallback seems to be a wrong name. I would expect to be something like: unsubmittedPeriodNotFoundCallback.

eiriko2303: Aha right, gonna fix it, thanks :D

src/pages/welcome-web/welcome-web.ts
```diff
@@ -177,7 +180,22 @@ export class WelcomeWebPage {
     });
   }
 
-  navigateToApproval(showPopup: boolean, rejectedNotFoundCallback?: Function) {
+  navigateToFirstUnsubmittedPeriod(showPopup: boolean, unsubmittedPeriodNotFoundCallback?: Function) {
+    if(this.periods) {
```
reg0: this piece of code is exactly the same in top-menu-mobile, couldn't we move it to popup.service? I know that all of these pop ups are duplicated all over the place, but I would be happy to finally keep them somewhere safe :)

PR #457, opened 2018-07-04
NT-1043 - Fixed bug that checkmark was not so responsive to clicks.

eiriko2303:
I also added some small design mods from Bodil (see NT-1059).


eiriko2303:
will do :)

PR #459, opened 2018-07-05
Nt 989 - firefox support

reg0:
I didn't fix one thing - hover on plus sign on "copy last day" button. The icon is being set in .ts and I didn't want to add extra complexity to handle mousenter and mouseleave events there. On the other hand, mobile version is not supposed to be viewed in Firefox on desktops so I don't think that's very important


reg0:
@eiriko2303 
I fixed most of the issues you reported;
- reports work fine now, I didn't have problem with dropdowns though, so please re-test and let me know how to reproduce it if it still happens
- I couldn't find any solution for tabbing problem, so I did a workaround for firefox by using markRelativeInput() on keyPressed event, but it's not 100% accurate, it skips comment icon and it doesn't advance to next row :/

eiriko2303:
Now it works great :) 👍 Nice job! The only thing that I dont understand is why firefox dont detect that I am choosing a customer. Have you that problem @reg0 ?

eiriko2303:
@reg0 Probably best to pull in the latest changes i did to the checkbox for better responsivness, to see if that solve the issue @noro5 mentioned :-)

reg0:
@eiriko2303 I merged dev, but it wasn't that
it turns out that ```@media(hover: hover)``` is not supported by FF, and I had to add some wacky media query to match FF as well ;)

@noro5 pls re-test

eiriko2303:
@reg0 Aha ok, great that you found a way for FF  too :-)

PR #470, opened 2018-07-09
NT-1055 Hours approved by others

sebdak:
move arrow to the right, and add blue text on hover


src/pages/approval/approval.scss
```diff
@@ -249,6 +253,9 @@
         display: block;
         &-clickable {
             cursor: pointer;
+            &:hover {
```
reg0: I think this header should remain blue even if not hovered? 
all other clickable/expandable elements in the app are like that

PR #471, opened 2018-07-10
NT-98: implement SSO

noro5:


src/app/app.component.ts
```diff
@@ -127,7 +127,18 @@ export class MyApp {
     this.keyboard$.subscribe(show => (this.showingKeyboard = show));
   }
 
+  private checkAdLogin() {
+    const startIndex = window.location.hash.indexOf("=");
```
reg0: this assumes that token is passed as a first parameter and there is some other one (ampersand has to be present), but I wouldn't rely on such assumptions

as discussed, you can do 
```
const startIndex = window.location.hash.indexOf("token=")+6;
let endIndex = window.location.hash.indexOf("&", startIndex);
endIndex = endIndex >= 0 ? endIndex : window.location.hash.length;
``` 
or 
```let token = window.location.hash.split("token=")[1].split('&')[0];```

File: src/pages/home/home.ts [hunk 2]
```diff
@@ -97,6 +103,37 @@ export class HomePage {
...
+    if (this.isWeb()) {
+      this.loginByBrowser();
+    } else {
+      this.loginByAdalPlugin();
+    }
+  }
+
+  public loginByBrowser() {
+    const callbackUrl = `${window.location.origin}/`;
+    this.storage.setItem('user_country', this.country);
```
reg0: why does this need to be stored here?

eiriko2303: I think that is because we dont get a value from the id token of which country table we should show the user. So we need the country value at login to now which countrycode database the user is going to have.

noro5: Yes, @eiriko2303 is right. Before I navigate to the SSO, I need to save the information from which country the user is, as this is not returned from Azure.

PR #475, opened 2018-07-18
Nt 1082 - Loading spinner for statistics page

mariusch:


src/components/statistics-toolbar/statistics-toolbar.ts
```diff
@@ -116,6 +119,7 @@ export class StatisticsToolbarComponent {
   }
 
   toggleChange(toggleValue: YearMonthState) {
+    this.ngRedux.dispatch(setShowSpinner(true));
```
reg0: usually it's not good to show the spinner without hiding it within the same function - we're not sure whether anything is connected to event emitter, and if it is, whether will it hide the spinner, so I would move those lines to the beginning of event handler

src/components/statistics-toolbar/statistics-toolbar.ts
```diff
@@ -125,6 +129,7 @@ export class StatisticsToolbarComponent {
   }
 
   selectChange(selectedValue) {
+    this.ngRedux.dispatch(setShowSpinner(true));
```
reg0: same here

src/pages/statistics/statistics.ts
```diff
@@ -83,6 +84,7 @@ export class StatisticsPage {
             this.ngRedux.dispatch(setAwayFromWorkEntriesForStatistics(entries.filter(entry => entry.awayFromWork === true)));
             this.ngRedux.dispatch(setNonBillableEntriesForStatistics(entries.filter(entry => entry.chargeable === false && entry.awayFromWork === false)));
             this.ngRedux.dispatch(setBillableEntriesForStatistics(entries.filter(entry => entry.chargeable === true)));
+            this.ngRedux.dispatch(setShowSpinner(false));
```
reg0: there are couple of async calls here, all of which may possibly fail, and spinner will be hidden only if they all succeed
please make sure that all uf this .subscribe() calls cannot fail, or if they can, then proper error handler should be added here to hide the spinner

sebdak:
@reg0 now it hides the spinner if the http req would fail with some error handling. not sure if this solution is what you were thinking of?

src/services/statistics.service.ts
```diff
@@ -39,20 +39,29 @@ export class StatisticsService {
             const arr = EntryForStatistics.mapFromApiArray(res.data);
             this.normTimeForPeriod = res.normTime;
             return arr;
+        })
+        .catch(err => {
```
reg0: wouldn't it work without it? it just rethrows the error, so I think it's not necessary

src/pages/statistics/statistics.ts
```diff
@@ -86,8 +86,15 @@ export class StatisticsPage {
...
             this.cdr.markForCheck();
+        }, err => {
+          console.log("error retrieving away from work project codes: " + err.message);
         });
+      }, err => {
+        console.log("error retrieving internal chargable tasks: " + err.message);
       });
+    }, err => {
+      console.log("error retrieving entries: " + err.message);
+      this.ngRedux.dispatch(setShowSpinner(false));
```
reg0: it should be probably repeated in each of the above ones

reg0: still though, setShowSpinner(true) is missing in the beginning of this function

reg0:
@sebdak in general, yes, but I think each subscribe should have it repeated

sebdak:
@reg0 we merged the away from work branch (that you just reviewed) into NT-76, and then pulled the changes into this branch. I think i've added proper error handling now so that the spinner gets turned off in case of eventual httpreq fail :)

PR #476, opened 2018-07-18
Nt 1071 - Away from work (statistics page)

mariusch:
Note:
* The real .svg's from Bodil are missing so we traced our own. They should be replaced next week when she's back from holiday.
* We made some small adjustments for the key number placement after feedback from Nitharshaan, so it don't look like they are table headings


src/components/away-from-work-key-numbers/away-from-work-key-numbers.component.ts
```diff
@@ -0,0 +1,21 @@
...
+  export class AwayFromWorkKeyNumbersComponent {
+
+    @Input() sickDays: number;
+    @Input() sickChildren: number;
+    @Input() holiday: number;
+
+    constructor() {}
+
+  isPluralDays(num): string {
+    return (num === 1 ? "day" : "days");
```
reg0: thumbs up for properly handling 0 ;)

src/models/enums/normTimeforCountries.enum.ts
```diff
@@ -0,0 +1,5 @@
+export enum NormTime {
+    NORWAY = 7.5,
+    SWEDEN = 8,
+    DENMARK = 7.24
```
reg0: 7:24h is not 7.24, but 7.4 :)

src/pages/away-from-work/away-from-work.ts
```diff
@@ -0,0 +1,64 @@
...
+  getKeyNumbers() {
+    let normtime = this.statService.getNormtimeHours();
+    let sickDays = (this.entries[0] !== undefined ? this.statService.getTotalHours(this.entries[0].data) / normtime : 0);
+    let sickChildren = (this.entries[1] !== undefined ? this.statService.getTotalHours(this.entries[1].data) / normtime : 0);
+    let holiday = (this.entries[2] !== undefined ? this.statService.getTotalHours(this.entries[2].data) / normtime : 0);
+    return [sickDays, sickChildren, holiday];
+  }
+
+ categorizeEntries(entries: Array<EntryForStatistics>) {
+    let sickDaysEntries: Array<EntryForStatistics> = entries.filter(entry => entry.projectNo === "05");
```
reg0: not sure if it would make sense here, but I don't like these hardcoded "magic numbers"
are we 100% sure they will not ever change? :)
maybe it'd be better to put them in enviroment.ts ?

PR #477, opened 2018-07-26
Nt 1083 - Refresh data in parent view when pop from child view

sebdak:


src/components/back-button/back-button.ts
```diff
@@ -10,11 +10,12 @@ export class BackButtonComponent {
   @Input() title: string;
   isWeb: boolean;
   
-  constructor(public navCtrl: NavController, private plt: Platform) {
+  constructor(public navCtrl: NavController, public navParams: NavParams, private plt: Platform) {
     this.isWeb = this.plt.is('core');
   }
 
   goBack() {
+    this.navParams.get("parentPage").refreshEntriesOnPagePop();
```
reg0: I don't like this, it assumes that there is such thing as "parentPage" and that it contains a certain method, and that might not be the case, especially in such generic and reusable component as "back-button" - this introduces very tight coupling which is always a bad thing

how about using ionic's component lifecycle, you could implement ```ionViewDidEnter()``` in StatisticsPage class and call refresh...() there? maybe some other of those functions, but I'd do it this way

see https://blog.ionicframework.com/navigating-lifecycle-events/ for reference

sebdak:
@reg0 Pushed the changes to for using ionViewWillEnter. The problem is that the first time we enter the statistics page, it makes two identical requests to the api. This is because it is the toolbar that emits an event to change the dateRange, and now we also do it in the ionViewWillEnter. It works, but we should find another solution

